PR #26, opened 2019-09-13
Shutdown callback support for all credentials providers

bretambrose:
* Shutdown callbacks for credentials providers can be configured from creation options structs; callback is made when all the provider's async resources are no longer in use (client bootstrap atm)
* aws_signer is removed and is conceptually a function pointer; all configurable signing state should go in a signing_config variant
* aws_signing_config_aws is now copied inside the sigv4 signing state and dynamically allocated state is copied as well; the passed in config does not need to persist for the duration of the signing process


include/aws/auth/credentials.h
```diff
@@ -90,13 +92,17 @@ struct aws_credentials_provider_chain_options {
 
 struct aws_credentials_provider_imds_options {
     struct aws_client_bootstrap *bootstrap;
+    aws_credentials_provider_bootstrap_released_fn *bootstrap_release_callback;
+    void *bootstrap_release_user_data;
```
graebm: Why is this the "bootstrap_release_callback"? Why not just have it be the "shutdown_complete_callback" and offer it on every single kind of credentials provider?

bretambrose: This is intended for CRTs to signal when they are no longer using resource objects.  The other providers don't use any resources so adding callbacks (and config) seems unnecessary since they wouldn't get used.

I started with shutdown_complete and that might have been the better choice.  There's no explicit shutdown on credentials providers though.  I wonder if resources_released might be clearer.

graebm: I would use something more generic than "bootstrap_released"
"resources_released" is fine

include/aws/auth/credentials.h
```diff
@@ -168,9 +188,7 @@ int aws_credentials_provider_get_credentials(
 AWS_AUTH_API
 struct aws_credentials_provider *aws_credentials_provider_new_static(
     struct aws_allocator *allocator,
-    struct aws_byte_cursor access_key_id,
-    struct aws_byte_cursor secret_access_key,
-    struct aws_byte_cursor session_token);
+    struct aws_credentials_provider_static_options *options);
```
graebm: nit: const struct *options

include/aws/auth/signing.h
```diff
@@ -98,11 +56,15 @@ int aws_signer_sign_request(
  * In all cases, the signing result will tell exactly what header and/or query params to add to the request
  * to become a fully-signed AWS http request.
  *
- * These restrictions can be relaxed, if necessary, in the future, but they don't unreasonable and
- * relaxing them adds non-trivial complexity.
+ * These restrictions can be relaxed, if necessary, in the future.
  */
 AWS_AUTH_API
-struct aws_signer *aws_signer_new_aws(struct aws_allocator *allocator);
+int aws_sign_request_aws(
```
graebm: Worth having a typedef to cover all signers?
`typedef int aws_signing_fn(.....);`

include/aws/auth/signing.h
```diff
@@ -98,11 +56,15 @@ int aws_signer_sign_request(
...
+    struct aws_allocator *allocator,
+    const struct aws_signable *signable,
+    const struct aws_signing_config_base *base_config,
```
ColdenCullen: This can be `aws_signing_config_aws` now!

source/aws_signing.c
```diff
@@ -249,24 +249,44 @@ static int s_append_signing_algorithm(enum aws_signing_algorithm algorithm, stru
...
+    if (state->region == NULL) {
+        goto on_error;
+    }
+    state->config.region = aws_byte_cursor_from_string(state->region);
+
+    state->service = aws_string_new_from_array(allocator, config->service.ptr, config->service.len);
+    if (state->service == NULL) {
+        goto on_error;
+    }
+    state->config.service = aws_byte_cursor_from_string(state->service);
```
ColdenCullen: See: whatever @graebm's thing is

source/credentials_utils.c
```diff
@@ -44,10 +44,14 @@ void aws_credentials_provider_init_base(
...
-    aws_atomic_store_int(&provider->shutting_down, 0);
+
     aws_atomic_store_int(&provider->ref_count, 1);
 }
 
-void aws_credentials_provider_shutdown_nil(struct aws_credentials_provider *provider) {
-    (void)provider;
+void aws_credentials_provider_invoke_shutdown_callback(struct aws_credentials_provider *provider) {
+    if (provider) {
+        if (provider->shutdown_options.shutdown_callback) {
```
ColdenCullen: [nit]
```c
if (provider && provider->shutdown_options.shutdown_callback) {
```

PR #33, opened 2019-10-29
Use security tokens

JonathanHenson:
Adds support for session based credentials to be used when they're present in the credentials.

I made up that test, because I just needed to make sure the signature was updated and that it was appropriately added to the signed output, and the official aws test suite does not include session tokens as part of the suite.


source/aws_signing.c
```diff
@@ -577,6 +578,18 @@ static int s_add_authorization_query_params(struct aws_signing_state_aws *state,
         goto done;
     }
 
+    /* X-Amz-Security-token */
+    if (state->config->credentials->session_token) {
```
bretambrose: This adds the param to what gets signed.  Does this need to be configurable (aws_signing_config boolean), defaulting to true?

source/signer.c
```diff
@@ -15,6 +15,7 @@
 
 #include <aws/auth/signer.h>
 
+#include <aws/auth/credentials.h>
```
bretambrose: Needed?

JonathanHenson: yes, it's dereferenced.

PR #35, opened 2019-11-06
Async Signer

ColdenCullen:
* `aws_signer_sign_request` is now async
* `aws_signing_config_aws` now uses a credentials provider instead of a raw credentials object

Also fixes small Mike-isms:
* Allow static credentials provider to accept an empty session token
* Add userdata to aws_should_sign_param_fn`

### DIFF
--- a/include/aws/auth/signing_config.h
+++ b/include/aws/auth/signing_config.h
@@ -23,7 +23,7 @@
 
 struct aws_credentials;
 
-typedef bool(aws_should_sign_param_fn)(const struct aws_byte_cursor *name);
+typedef bool(aws_should_sign_param_fn)(const struct aws_byte_cursor *name, void *userdata);
 
 /*
  * A primitive RTTI indicator for signing configuration structs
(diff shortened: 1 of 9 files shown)


include/aws/auth/signing_config.h
```diff
@@ -23,7 +23,7 @@
 
 struct aws_credentials;
 
-typedef bool(aws_should_sign_param_fn)(const struct aws_byte_cursor *name);
+typedef bool(aws_should_sign_param_fn)(const struct aws_byte_cursor *name, void *userdata);
```
graebm: trivial: we usually spell it `user_data`

source/aws_profile.c
```diff
@@ -556,10 +556,8 @@ static int s_profile_merge(struct aws_profile *dest_profile, const struct aws_pr
                 return AWS_OP_ERR;
             }
 
-            struct aws_byte_cursor empty_value;
-            AWS_ZERO_STRUCT(empty_value);
-
             struct aws_byte_cursor property_name = aws_byte_cursor_from_string(dest_key);
+            static struct aws_byte_cursor empty_value = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("");
             dest_property = aws_profile_property_new(dest_profile->allocator, &property_name, &empty_value);
```
graebm: weird.
you said you were working around  a crash in `aws_profile`
we should probably just fix that so it can take cursor with NULL ptr

ColdenCullen: Determined this is just an issue with my environment. Reverting for now, will come back with a better fix later.

PR #38, opened 2019-11-08
Don't add the sha256 header if sign_body is off

ColdenCullen:

### DIFF
--- a/source/aws_signing.c
+++ b/source/aws_signing.c
@@ -1111,7 +1111,7 @@ static int s_append_canonical_payload_hash(struct aws_signing_state_aws *state) 
     /*
      * Add the payload hash header to the result if necessary
      */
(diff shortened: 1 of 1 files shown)


bretambrose:
> This is incorrect:
> 
> You add an empty signature if sign_body is off:
> 
> it should be set to the value:
> e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855

This is the signing result not the canonical request

JonathanHenson:
It’s always added to the signing result, at least in header signing mode:
https://github.com/aws/aws-sdk-cpp/blob/master/aws-cpp-sdk-core/source/auth/AWSAuthSigner.cpp#L236

ColdenCullen:
thank mr henos

PR #51, opened 2019-12-31
Fixed parallel test runs using the same file name and stomping on eac…

JonathanHenson:
…h other.

### DIFF
--- a/tests/credentials_provider_sts_tests.c
+++ b/tests/credentials_provider_sts_tests.c
@@ -594,23 +594,15 @@ static int s_credentials_provider_sts_from_profile_config_succeeds_fn(struct aws
 
     struct aws_string *config_contents = aws_string_new_from_c_str(allocator, s_soure_profile_config_file);
 
-    char config_file_storage[64] = {0};
-    char creds_file_storage[64] = {0};
-
-    struct aws_byte_cursor config_file_cur =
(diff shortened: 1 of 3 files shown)


tests/credentials_provider_sts_tests.c
```diff
@@ -593,12 +593,24 @@ static int s_credentials_provider_sts_from_profile_config_succeeds_fn(struct aws
...
+
+    char config_file_storage[64] = {0};
+    char creds_file_storage[64] = {0};
+
+    struct aws_byte_cursor config_file_cur =
+        aws_create_process_unique_file_name(config_file_storage, sizeof(config_file_storage));
+    struct aws_byte_cursor creds_file_cur =
+        aws_create_process_unique_file_name(creds_file_storage, sizeof(creds_file_storage));
+
+    struct aws_string *creds_file_str = aws_string_new_from_array(allocator, creds_file_cur.ptr, creds_file_cur.len);
```
graebm: trivial: why not just have unique_file_name() return an `aws_string`?
Instead of needing to declare storage, and a byte_cursor, and also create an aws_string?

JonathanHenson: well we don't create a string everywhere. some places we just pass a cursor. Which do you PREFER? I'll just do that.

PR #57, opened 2020-02-26
Do not export cJSON symbols in DLLs

kankri:
If you use aws-c-auth in your code which you compile into a DLL, it gets 74 exported internal aws-c-auth symbols.

```
>dumpbin.exe /exports foo.dll
[...]
    ordinal hint RVA      name

          1    0 00001A7D _cJSON_AddArrayToObject@8 = @ILT+2680(_cJSON_AddArrayToObject@8)
[...]
         74   49 000047D2 _cJSON_malloc@4 = @ILT+14285(_cJSON_malloc@4)
```

Steps to reproduce:

   1. build aws-c-auth
   2. use aws-c-auth in your DLL
       ```
       find_package(aws-c-auth REQUIRED)
	   add_library(foo SHARED)
       ```

The reason for the problem is that cJSON by default exports all its symbols into any DLL it is used in.

This PR fixes the place in aws-c-auth where cJSON is used to suppress exporting its symbols (there is a [similar problem](https://github.com/awslabs/aws-crt-cpp/pull/100) in aws-crt-cpp).

---


bretambrose:
Cloned and merged your PR.  Thank you for finding and fixing this.

PR #59, opened 2020-02-26
Support ECS Creds Provider

singku-china:
*Issue #, if available:*

*Description of changes:*


include/aws/auth/credentials.h
```diff
@@ -157,6 +157,35 @@ struct aws_credentials_provider_imds_options {
...
+ * from endpoint are needed.
+ */
+struct aws_credentials_provider_ecs_options {
+    struct aws_credentials_provider_shutdown_options shutdown_options;
+    struct aws_client_bootstrap *bootstrap;
+
+    struct aws_byte_cursor host;
+    struct aws_byte_cursor path_and_query;
+    struct aws_byte_cursor auth_token;
+    /* it is also used to deterine the port: 443 or 80 */
```
graebm: nit: "determine"

source/credentials.c
```diff
@@ -199,8 +207,17 @@ struct aws_credentials_provider *aws_credentials_provider_new_chain_default(
 
     environment_provider = aws_credentials_provider_new_environment(allocator, &environment_options);
 
+    bool success = true;
+    struct aws_string *ecs_relative_uri = NULL;
+    aws_get_environment_value(allocator, s_ecs_creds_env_relative_uri, &ecs_relative_uri);
+    struct aws_string *ecs_full_uri = NULL;
+    aws_get_environment_value(allocator, s_ecs_creds_env_full_uri, &ecs_full_uri);
+    struct aws_string *ec2_imds_disable = NULL;
+    aws_get_environment_value(allocator, s_ec2_creds_env_disable, &ec2_imds_disable);
```
graebm: need to check return code of `aws_get_environment_value()`

source/credentials.c
```diff
@@ -199,8 +207,17 @@ struct aws_credentials_provider *aws_credentials_provider_new_chain_default(
 
     environment_provider = aws_credentials_provider_new_environment(allocator, &environment_options);
 
+    bool success = true;
```
graebm: It would be simpler (and less error prone) if `bool success = false` by default, and we only set it true in one place.

instead of needing to remember to set it false in the many places an error might occur

source/credentials.c
```diff
@@ -213,34 +230,82 @@ struct aws_credentials_provider *aws_credentials_provider_new_chain_default(
...
-    if (imds_provider == NULL) {
-        goto on_error;
-    }
+    if (ecs_relative_uri && ecs_relative_uri->len) {
+        struct aws_credentials_provider_ecs_options ecs_options;
+        AWS_ZERO_STRUCT(ecs_options);
+        ecs_options.bootstrap = options->bootstrap;
+        ecs_options.host = aws_byte_cursor_from_string(s_ecs_host);
+        ecs_options.path_and_query = aws_byte_cursor_from_string(ecs_relative_uri);
+        ecs_options.use_tls = false;
```
graebm: Don't change this, you're fitting in with the style of the code around you. I just wanted to share, in case you didn't know, we can do pretty things like this in C:
```
struct aws_credentials_provider_ecs_options ecs_options = {
    .bootstrap = options->bootstrap,
    .host = aws_byte_cursor_from_string(s_ecs_host),
    .path_and_query = aws_byte_cursor_from_string(ecs_relative_uri),
    .use_tls = false,
};
```

source/credentials.c
```diff
@@ -213,34 +230,82 @@ struct aws_credentials_provider *aws_credentials_provider_new_chain_default(
...
+        struct aws_uri uri;
+        struct aws_byte_cursor uri_cstr = aws_byte_cursor_from_string(ecs_full_uri);
+        if (AWS_OP_ERR == aws_uri_init_parse(&uri, allocator, &uri_cstr)) {
+            success = false;
+            goto on_ret;
+        }
 
-    providers[index] = imds_provider;
+        struct aws_string *ecs_token = NULL;
+        aws_get_environment_value(allocator, s_ecs_creds_env_token, &ecs_token);
```
graebm: need to check return value of `aws_get_environment_value()`

source/credentials.c
```diff
@@ -213,34 +230,82 @@ struct aws_credentials_provider *aws_credentials_provider_new_chain_default(
...
+        struct aws_credentials_provider_ecs_options ecs_options;
+        AWS_ZERO_STRUCT(ecs_options);
+        ecs_options.bootstrap = options->bootstrap;
+        ecs_options.host = uri.host_name;
+        ecs_options.path_and_query = uri.path_and_query;
+        ecs_options.use_tls = aws_byte_cursor_eq_c_str_ignore_case(&(uri.scheme), "HTTPS");
+        if (ecs_token->len) {
```
graebm: never checked if ecs_token was NULL

source/credentials.c
```diff
@@ -199,8 +207,17 @@ struct aws_credentials_provider *aws_credentials_provider_new_chain_default(
...
+    aws_get_environment_value(allocator, s_ecs_creds_env_relative_uri, &ecs_relative_uri);
+    struct aws_string *ecs_full_uri = NULL;
+    aws_get_environment_value(allocator, s_ecs_creds_env_full_uri, &ecs_full_uri);
+    struct aws_string *ec2_imds_disable = NULL;
+    aws_get_environment_value(allocator, s_ec2_creds_env_disable, &ec2_imds_disable);
+
     if (environment_provider == NULL) {
-        goto on_error;
+        success = false;
+        goto on_ret;
```
graebm: move this check up to happen immediately after the `aws_credentials_provider_new_environment()` call. It's weird having the check later after doing a bunch of unrelated stuff.

It's more common to put things at the top of the function that will get set later on (in this case: `success`, and the 3 new ecs strings)

source/credentials.c
```diff
@@ -213,34 +230,82 @@ struct aws_credentials_provider *aws_credentials_provider_new_chain_default(
...
+        ecs_options.bootstrap = options->bootstrap;
+        ecs_options.host = aws_byte_cursor_from_string(s_ecs_host);
+        ecs_options.path_and_query = aws_byte_cursor_from_string(ecs_relative_uri);
+        ecs_options.use_tls = false;
+        ecs_provider = aws_credentials_provider_new_ecs(allocator, &ecs_options);
+        if (ecs_provider == NULL) {
+            success = false;
+            goto on_ret;
+        }
+        providers[index] = ecs_provider;
```
graebm: why isn't this `providers[index++] = ecs_provider` ?

source/credentials.c
```diff
@@ -213,34 +230,82 @@ struct aws_credentials_provider *aws_credentials_provider_new_chain_default(
...
+        }
+        providers[index] = imds_provider;
+    }
 
     AWS_FATAL_ASSERT(index < AWS_ARRAY_SIZE(providers));
 
     struct aws_credentials_provider_chain_options chain_options;
     AWS_ZERO_STRUCT(chain_options);
-    chain_options.provider_count = index;
+    chain_options.provider_count = index + 1;
```
graebm: I think we should have just been doing `providers[index++] = x;` everywhere
which means you could revert this change,
but we should modify the assertion above to be `AWS_FATAL_ASSERT(index <= AWS_ARRAY_SIZE(providers));` instead of just `<`

source/credentials.c
singku-china: Agreed

source/credentials.c
singku-china: Thought of in the following code, testing if the string pointers are `NULL` and if the `len` vars are non zero show cover the return of this function.

source/credentials.c
singku-china: Yes, followed into the sense of `define it where you use it`, then due to the `goto` stuff, I have to move them up but was not in the correct place.

source/credentials.c
singku-china: I had the same question and think your idea is the correct way to do, but not quit sure why was this way originally. (originally the first two are p[index++] = provider, but the last one didn't increase index)

source/credentials.c
singku-china: good catch

source/credentials.c
singku-china: let's change it

source/credentials.c
graebm: `define it where you use it` is a good idiom in OTHER languages that automatically destroy objects that go out of scope. The ugly habit of defining everything at the top of a function came from C, where objects don't clean themselves up, so we have to use `goto` and jump to code that cleans everything up, and the things we need to clean up must be defined before our first `goto`.

You should still "define it where you use it" for things like random ints and byte_cursors that don't need clean-up. Only define variables early if you need to clean them up later.

graebm: Actually, earlier versions of C didn't even let you define variables except at the top of the function, so I guess that's really where the bad habit came from

source/credentials.c
graebm: If `aws_get_environment_value()` fails, it's because the user **does** have the environment variable set, but we ran out of memory and couldn't return the string. So we'd be doing the wrong thing to proceed as if the environment variable wasn't set

source/credentials.c
graebm: Talked with Jonathan who put this index variable in to the function. He agrees we should just be `++`ing it everywhere

source/credentials.c
singku-china: Got your point, I'll change this.

source/credentials_provider_ecs.c
```diff
@@ -0,0 +1,636 @@
...
+        goto done;
+    }
+
+    credentials = aws_credentials_new_from_cursors(
+        allocator, &access_key_id_cursor, &secret_access_key_cursor, &session_token_cursor);
+
+    if (creds_expiration_cursor.len != 0) {
+        struct aws_date_time expiration;
+        aws_date_time_init_from_str_cursor(&expiration, &creds_expiration_cursor, AWS_DATE_FORMAT_ISO_8601);
+        credentials->expiration_timepoint_seconds = expiration.timestamp;
```
graebm: don't access the internal members of `aws_data_time` directly, use the public API. I think you want `aws_date_time_as_epoch_secs()`?
If API functions exist for a struct, you should consider its member variables private. Yes, we declared `struct aws_date_time` in a header, but only did that so it can be used without requiring a heap allocation.

The only exceptions where it's ok to access the member variables are:
- `aws_byte_cursor`
- "options" structs like `aws_credentials_provider_ecs_options` that have no API functions.
- an implementation can touch its own members of course

source/credentials_provider_ecs.c
```diff
@@ -0,0 +1,636 @@
...
+    /* Try to build credentials from whatever, if anything, was in the result */
+    struct aws_credentials *credentials =
+        s_parse_credentials_from_ecs_document(ecs_user_data->allocator, &ecs_user_data->current_result);
+
+    /* pass the credentials back */
+    ecs_user_data->original_callback(credentials, ecs_user_data->original_user_data);
+    if (credentials != NULL) {
+        AWS_LOGF_INFO(
+            AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+            "(id=%p) ECS credentials provider successfully queried instance role credentials",
```
graebm: trivial: log these messages before invoking the callback

source/credentials_provider_ecs.c
```diff
@@ -0,0 +1,636 @@
...
+    void *user_data) {
+
+    (void)stream;
+
+    struct aws_credentials_provider_ecs_user_data *ecs_user_data = user_data;
+    struct aws_credentials_provider_ecs_impl *impl = ecs_user_data->ecs_provider->impl;
+
+    AWS_LOGF_DEBUG(
+        AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+        "(id=%p) ECS credentials provider received %zu response bytes",
```
graebm: trivial/debatable: this seems like more of a TRACE level message?

source/credentials_provider_ecs.c
```diff
@@ -0,0 +1,636 @@
...
+        data->len);
+
+    if (data->len + ecs_user_data->current_result.len > ECS_RESPONSE_SIZE_LIMIT) {
+        impl->function_table->aws_http_connection_close(ecs_user_data->connection);
+        AWS_LOGF_ERROR(
+            AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+            "(id=%p) ECS credentials provider query response exceeded maximum allowed length",
+            (void *)ecs_user_data->ecs_provider);
+
+        return AWS_OP_ERR;
```
graebm: need to aws_raise_error()

source/credentials_provider_ecs.c
```diff
@@ -0,0 +1,636 @@
...
+    }
+
+    if (aws_byte_buf_append_dynamic(&ecs_user_data->current_result, data)) {
+        impl->function_table->aws_http_connection_close(ecs_user_data->connection);
+        AWS_LOGF_ERROR(
+            AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+            "(id=%p) ECS credentials provider query error appending response",
+            (void *)ecs_user_data->ecs_provider);
+
+        return AWS_OP_ERR;
```
graebm: need to aws_raise_error()

source/credentials_provider_ecs.c
```diff
@@ -0,0 +1,636 @@
...
+    struct aws_credentials_provider_ecs_impl *impl = ecs_user_data->ecs_provider->impl;
+    impl->function_table->aws_http_stream_release(stream);
+
+    /*
+     * On anything other than a 200, nullify the response and pretend there was
+     * an error
+     */
+    if (ecs_user_data->status_code != 200) {
+        ecs_user_data->current_result.len = 0;
+    }
```
graebm: probably also want to nullify the answer if `error_code` is non-zero. Currently `error_code` is ignored

source/credentials_provider_ecs.c
```diff
@@ -0,0 +1,636 @@
...
+
+    impl->connection_manager = impl->function_table->aws_http_connection_manager_new(allocator, &manager_options);
+    if (impl->connection_manager == NULL) {
+        goto on_error;
+    }
+    if (options->auth_token.len != 0) {
+        impl->auth_token = aws_string_new_from_array(allocator, options->auth_token.ptr, options->auth_token.len);
+    }
+    impl->path_and_query =
+        aws_string_new_from_array(allocator, options->path_and_query.ptr, options->path_and_query.len);
```
graebm: need to check results of `aws_string_new_xxx()`
could fail if we ran out of memory

source/credentials.c
```diff
@@ -213,56 +231,117 @@ struct aws_credentials_provider *aws_credentials_provider_new_chain_default(
...
-    struct aws_credentials_provider_imds_options imds_options;
-    AWS_ZERO_STRUCT(imds_options);
-    imds_options.bootstrap = options->bootstrap;
-    imds_provider = aws_credentials_provider_new_imds(allocator, &imds_options);
-    if (imds_provider == NULL) {
-        goto on_error;
+    if (aws_get_environment_value(allocator, s_ecs_creds_env_relative_uri, &ecs_relative_uri) != AWS_OP_SUCCESS ||
+        aws_get_environment_value(allocator, s_ecs_creds_env_full_uri, &ecs_full_uri) != AWS_OP_SUCCESS ||
+        aws_get_environment_value(allocator, s_ec2_creds_env_disable, &ec2_imds_disable) != AWS_OP_SUCCESS) {
+        AWS_LOGF_FATAL(
```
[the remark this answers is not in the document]
singku-china: Shall we fatal and abort when out of memory?

source/credentials_provider_ecs.c
singku-china: What kind of error shall we raise is appropriate?

source/credentials.c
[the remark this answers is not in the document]
graebm: No, we just treat it as a normal error.
Clean up whatever we were in the middle of and return AWS_OP_ERR

source/credentials_provider_ecs.c
```diff
@@ -0,0 +1,636 @@
...
+    cJSON *document_root = NULL;
+
+    struct aws_byte_cursor null_terminator_cursor = aws_byte_cursor_from_string(s_empty_empty_string);
+    if (aws_byte_buf_append_dynamic(document, &null_terminator_cursor)) {
+        goto done;
+    }
+
+    document_root = cJSON_Parse((const char *)document->buffer);
+    if (document_root == NULL) {
+        goto done;
```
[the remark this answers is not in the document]
graebm: still need to `aws_raise_error()` when these cJSON_() files don't return what you want

source/credentials_provider_ecs.c
```diff
@@ -0,0 +1,653 @@
...
+    if (creds_expiration_cursor.len != 0) {
+        struct aws_date_time expiration;
+        if (aws_date_time_init_from_str_cursor(&expiration, &creds_expiration_cursor, AWS_DATE_FORMAT_ISO_8601) ==
+            AWS_OP_ERR) {
+            AWS_LOGF_ERROR(
+                AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+                "Expiration in ECS response Json document is not a valid ISO_8601 date string.");
+            goto done;
+        }
+        credentials->expiration_timepoint_seconds = (uint64_t)aws_date_time_as_epoch_secs(&expiration);
```
graebm: need to immediately check `credentials` to see if it's NULL
we're referencing `credentials->expiration_timepoint_seconds` later, which would segfault if `credentials` was NULL

source/credentials_provider_ecs.c
```diff
@@ -0,0 +1,653 @@
...
+        allocator, &access_key_id_cursor, &secret_access_key_cursor, &session_token_cursor);
+
+    if (creds_expiration_cursor.len != 0) {
+        struct aws_date_time expiration;
+        if (aws_date_time_init_from_str_cursor(&expiration, &creds_expiration_cursor, AWS_DATE_FORMAT_ISO_8601) ==
+            AWS_OP_ERR) {
+            AWS_LOGF_ERROR(
+                AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+                "Expiration in ECS response Json document is not a valid ISO_8601 date string.");
+            goto done;
```
graebm: if this fails, we need to delete `credentials` and return NULL

source/credentials_provider_ecs.c
graebm: Seems like we'll need some new errors, maybe something like `AWS_AUTH_CREDENTIALS_PARSE_ERROR` ?

source/credentials_provider_ecs.c
```diff
@@ -0,0 +1,653 @@
...
+    }
+
+    impl->connection_manager = impl->function_table->aws_http_connection_manager_new(allocator, &manager_options);
+    if (impl->connection_manager == NULL) {
+        goto on_error;
+    }
+    if (options->auth_token.len != 0) {
+        impl->auth_token = aws_string_new_from_array(allocator, options->auth_token.ptr, options->auth_token.len);
+        if (impl->auth_token == NULL) {
+            aws_raise_error(AWS_ERROR_OOM);
```
graebm: trivial: So, the rule is: if an aws_XYZ() function returns a NULL or AWS_OP_ERR, then that function has already set the error code, you don't need to assume that it failed due to OOM and call aws_raise_error() again.

You only need to aws_raise_error() if something went wrong OUTSIDE a call to aws_XYZ()

source/credentials.c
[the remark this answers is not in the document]
singku-china: ok, here the return is provider, AWS_OP_ERR should has already been logged in the inner call, we can log and return NULL.

source/credentials.c
```diff
@@ -213,56 +231,117 @@ struct aws_credentials_provider *aws_credentials_provider_new_chain_default(
         providers[index++] = profile_provider;
     }
 
-    struct aws_credentials_provider_imds_options imds_options;
-    AWS_ZERO_STRUCT(imds_options);
-    imds_options.bootstrap = options->bootstrap;
-    imds_provider = aws_credentials_provider_new_imds(allocator, &imds_options);
-    if (imds_provider == NULL) {
-        goto on_error;
+    if (aws_get_environment_value(allocator, s_ecs_creds_env_relative_uri, &ecs_relative_uri) != AWS_OP_SUCCESS ||
```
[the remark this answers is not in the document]
singku-china: This is a good idea. I will do that

source/credentials_provider_ecs.c
```diff
@@ -0,0 +1,653 @@
...
+    void *user_data) {
+
+    struct aws_credentials_provider_ecs_user_data *wrapped_user_data =
+        aws_mem_calloc(ecs_provider->allocator, 1, sizeof(struct aws_credentials_provider_ecs_user_data));
+    if (wrapped_user_data == NULL) {
+        goto on_error;
+    }
+
+    wrapped_user_data->allocator = ecs_provider->allocator;
+    wrapped_user_data->ecs_provider = ecs_provider;
```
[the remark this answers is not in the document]
singku-china: I can change it all together in the IMDS secure PR.

source/credentials_provider_ecs.c
singku-china: What about AWS_ERROR_SHORT_BUFFER?

graebm: I think AWS_ERROR_SHORT_BUFFER would be much more confusing to the user than something like PARSE_ERROR

source/credentials_provider_ecs.c
```diff
@@ -0,0 +1,667 @@
...
+            credentials = NULL;
+            goto parse_error;
+        }
+        credentials->expiration_timepoint_seconds = (uint64_t)aws_date_time_as_epoch_secs(&expiration);
+    }
+
+parse_error:
+    aws_raise_error(AWS_AUTH_PROVIDER_PARSER_UNEXPECTED_RESPONSE);
+
+done:
```
graebm: trivial: maybe throw in a `goto done;` once we know it's all successful, so that we skip over the `parse_error:` section of the code when everything's going ok.

It's not technically wrong to aws_raise_error() when you're returning a successful result in the end, but it's a little weird

tests/credentials_tests.c
```diff
@@ -572,6 +572,8 @@ static int s_cached_credentials_provider_queued_async_test(struct aws_allocator
         0);
 
     aws_credentials_provider_release(cached_provider);
+    ASSERT_TRUE(aws_atomic_load_int(&(cached_provider->ref_count)) == 0);
```
bretambrose: I think these can go away now as they're not going to be reliably true due to async shutdown.

PR #60, opened 2020-03-03
Properly use and separate high res and system time

bretambrose:
* Correct math that involved system clock calculations being applied to high res timepoints.
* Added additional mocking so that cached credentials provider could be properly tested.
* Update tests appropriately.


include/aws/auth/credentials.h
```diff
@@ -140,7 +140,8 @@ struct aws_credentials_provider_cached_options {
     struct aws_credentials_provider_shutdown_options shutdown_options;
     struct aws_credentials_provider *source;
     uint64_t refresh_time_in_milliseconds;
-    aws_io_clock_fn *clock_fn;
+    aws_io_clock_fn *high_res_clock_fn;
+    aws_io_clock_fn *system_clock_fn;
```
graebm: /* For mocking, leave NULL otherwise */

PR #61, opened 2020-03-05
Fix Memory Leak in Test

singku-china:
*Issue #, if available:*
Following scripts will cache a memory leak caused by test `cached_credentials_provider_queued_async_test`
```
#!/bin/sh
while [ $? == 0 ]
do
    ./aws-c-auth-tests cached_credentials_provider_queued_async_test
done
```
The leak happens because:
- Mocked credentials provider will create a worker thread to process the queries.
- aws_credentials_query will ref cnt incr credentials provider, here is mocked credentials provider.
- Cached credentials provider's source is mocked provider.
- When test finished, cached provider got released in main thread and will wait its shutdown callback to be finished in `s_aws_wait_for_provider_shutdown_callback();` at main thread.
https://github.com/awslabs/aws-c-auth/blob/543e3764d7fa4f1228ed630d1aa40e4a0de396d6/tests/credentials_tests.c#L574-L576
- Well cached provider's shutdown callback is invoked when mock provider get destroyed. **here is the trick and important part**. 
  - Mocked provider's shutdown callback is invoked in its destroy function. https://github.com/awslabs/aws-c-auth/blob/543e3764d7fa4f1228ed630d1aa40e4a0de396d6/tests/credentials_provider_utils.c#L232
  - This call back is https://github.com/awslabs/aws-c-auth/blob/543e3764d7fa4f1228ed630d1aa40e4a0de396d6/source/credentials_provider_cached.c#L323
  - Which is cached providers' concrete destroy function, in this function, it will invoke it's own shutdown callback, which is https://github.com/awslabs/aws-c-auth/blob/543e3764d7fa4f1228ed630d1aa40e4a0de396d6/tests/credentials_tests.c#L515
  - This callback happens before joining worker thread and will wake up main thread from `s_aws_wait_for_provider_shutdown_callback();`
- In this case, main thread continues and goto mem check in test finish, but if worker thread is still processing queries and not yet destroyed mock provider (because query will hold mock provider), mem check will report leak.

Apart from this leak, there is dead lock race condition in mock provider's destroy function.
- Worker thread will first acquire controller's mutex before doing its job. https://github.com/awslabs/aws-c-auth/blob/543e3764d7fa4f1228ed630d1aa40e4a0de396d6/tests/credentials_provider_utils.c#L264
- When doing its job, it may call query_clean_up, which will de ref cnt mock provider, and potentially call mock provider's destroy function if ref cnt becomes zero.https://github.com/awslabs/aws-c-auth/blob/543e3764d7fa4f1228ed630d1aa40e4a0de396d6/tests/credentials_provider_utils.c#L298
- Well in destroy function, it tries to acquire lock again. https://github.com/awslabs/aws-c-auth/blob/543e3764d7fa4f1228ed630d1aa40e4a0de396d6/tests/credentials_provider_utils.c#L234

*Description of changes:*
First things first, debugging cached creds provider with async mock is a nightmare. 

- Changed mock provider's callback to happen after thread join, so that making sure main thread will wait worker thread to finish first.
- Removed provider reference in query so that clean up query won't cause dead lock. This is still in question though because currently provider in query is not used anywhere, but I don't know the purpose why we have it.

### DIFF
--- a/include/aws/auth/private/credentials_utils.h
+++ b/include/aws/auth/private/credentials_utils.h
@@ -84,7 +84,6 @@ AWS_EXTERN_C_BEGIN
 AWS_AUTH_API
 void aws_credentials_query_init(
     struct aws_credentials_query *query,
-    struct aws_credentials_provider *provider,
     aws_on_get_credentials_callback_fn *callback,
     void *user_data);
 
--- a/source/credentials_provider_cached.c
+++ b/source/credentials_provider_cached.c
@@ -166,7 +166,7 @@ static int s_cached_credentials_provider_get_credentials_async(
         struct aws_credentials_query *query =
             aws_mem_acquire(provider->allocator, sizeof(struct aws_credentials_query));
         if (query != NULL) {
-            aws_credentials_query_init(query, provider, callback, user_data);
+            aws_credentials_query_init(query, callback, user_data);
             should_submit_query = aws_linked_list_empty(&impl->pending_queries);
             aws_linked_list_push_back(&impl->pending_queries, &query->node);
         } else {
--- a/source/credentials_utils.c
+++ b/source/credentials_utils.c
@@ -17,22 +17,15 @@
 
 void aws_credentials_query_init(
     struct aws_credentials_query *query,
-    struct aws_credentials_provider *provider,
     aws_on_get_credentials_callback_fn *callback,
     void *user_data) {
     AWS_ZERO_STRUCT(*query);
-
-    query->provider = provider;
     query->user_data = user_data;
     query->callback = callback;
-
-    aws_credentials_provider_acquire(provider);
 }
 
 void aws_credentials_query_clean_up(struct aws_credentials_query *query) {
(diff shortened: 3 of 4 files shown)


singku-china:
But if a cached provider itself acquires source provider, as long as the cached provider still lives, the source provider won't go away right no matter the queries right?

bretambrose:
> But if a cached provider itself acquires source provider, as long as the cached provider still lives, the source provider won't go away right no matter the queries right?

(I think we resolved this on chat, but just posting reply for documentation)

The source provider will persist yes, but if the cached provider goes away, pending queries to the source provider will yield a crash when they complete (the query's userdata value is the old pointer to the caching provider which is now garbage).  Essentially, this helps prevent accidental misuse where someone makes a query and then doesn't wait for it to complete before dropping their reference to the caching provider.  More generally, I think it's best if asynchronous execution paths always clean up safely rather than relying on the user to keep explicit track of every invocation they make.

PR #63, opened 2020-03-11
Support IMDS Credentials Provider Secure Request.

singku-china:
*Issue #, if available:*
In addition to existing two requests of IMDSV2, IMDS V2 requires an additional token request (using Http GET method) prior to these two existing requests. Then in the existing requests, we need to add a new token header with the value acquired from the first request.

By default SDK should try using IMDS V2, but if customer set token not required, then IMDSV1 should be tried first. SDK should also switch automatically between V1 and V2 based on the service response.

IMDSV1 -> first role name request get 401 unauthorized -> IMDSV2
IMDSV2 -> first token request get get non 400 error -> IMDSV1
 
*Description of changes:*


source/credentials_provider_imds.c
```diff
@@ -333,9 +386,45 @@ static void s_imds_on_stream_complete_fn(struct aws_http_stream *stream, int err
...
+
+        /*
+         * Other than that, if meets any error, then token is not required,
+         * we should fall back to insecure request. Otherwise, we should use
+         * token in following requests.
+         */
+        if (imds_user_data->status_code != 200 || imds_user_data->current_result.len == 0) {
+            impl->token_required = false;
+        } else {
+            imds_user_data->token = aws_byte_cursor_from_buf(&(imds_user_data->current_result));
```
bretambrose: Doesn't current_result get cleared?  Token needs to be used for up to two more queries right?

source/credentials_provider_imds.c
```diff
@@ -333,9 +386,45 @@ static void s_imds_on_stream_complete_fn(struct aws_http_stream *stream, int err
...
+            return;
+        }
+
+        /*
+         * Other than that, if meets any error, then token is not required,
+         * we should fall back to insecure request. Otherwise, we should use
+         * token in following requests.
+         */
+        if (imds_user_data->status_code != 200 || imds_user_data->current_result.len == 0) {
+            impl->token_required = false;
```
bretambrose: I don't think we should be modifying state on the provider in the middle of an async call.  In fact, it seems like token_required should be a part of the query (imds_user_data).  The provider's copy should be a read-only default.

source/credentials_provider_imds.c
```diff
@@ -393,11 +486,37 @@ static int s_make_imds_http_query(
         goto on_error;
     }
 
-    if (aws_http_message_set_request_path(request, *uri)) {
-        goto on_error;
+    struct aws_credentials_provider_imds_impl *impl = imds_user_data->imds_provider->impl;
+
+    if (imds_user_data->query_state == AWS_IMDS_QS_TOKEN) {
```
bretambrose: Feels like this new block could go in a separate token-oriented function.

singku-china: Yes, it's better

source/credentials_provider_imds.c
```diff
@@ -162,33 +170,47 @@ static struct aws_credentials *s_parse_credentials_from_imds_document(
     struct aws_credentials *credentials = NULL;
     cJSON *document_root = NULL;
 
-    struct aws_byte_cursor null_terminator_cursor = aws_byte_cursor_from_string(s_empty_empty_string);
+    struct aws_byte_cursor null_terminator_cursor = aws_byte_cursor_from_string(s_empty_string);
```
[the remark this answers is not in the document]
singku-china: I will make changes and change ECS one as well.

source/credentials_provider_imds.c
```diff
@@ -322,8 +386,48 @@ static int s_imds_on_incoming_headers_fn(
     return AWS_OP_SUCCESS;
 }
 
+static void s_imds_query_token(struct aws_credentials_provider_imds_user_data *aws_credentials_provider_imds_user_data);
+static void s_imds_query_instance_role_name(struct aws_credentials_provider_imds_user_data *imds_user_data);
 static void s_imds_query_instance_role_credentials(struct aws_credentials_provider_imds_user_data *imds_user_data);
 
+static bool s_isspace(unsigned char c) {
```
bretambrose: Very minor - prefer the sized types when its our own code (uint8_t)

source/credentials_provider_imds.c
```diff
@@ -322,8 +386,48 @@ static int s_imds_on_incoming_headers_fn(
...
+    return isspace((int)c) != 0;
+}
+
+static int s_imds_on_token_response(void *user_data) {
+
+    struct aws_credentials_provider_imds_user_data *imds_user_data = user_data;
+
+    /* Gets 400 means token is required but the request itself failed. */
+    if (imds_user_data->status_code == 400) {
+        return -1;
```
bretambrose: AWS_OP_SUCCESS/AWS_OP_ERR here and a number of other places

source/credentials_provider_imds.c
```diff
@@ -333,9 +437,27 @@ static void s_imds_on_stream_complete_fn(struct aws_http_stream *stream, int err
     struct aws_credentials_provider_imds_impl *impl = imds_user_data->imds_provider->impl;
     impl->function_table->aws_http_stream_release(stream);
 
+    // Token requst finished
+    if (imds_user_data->query_state == AWS_IMDS_QS_TOKEN) {
+        if (s_imds_on_token_response(user_data) != 0) {
```
bretambrose: Go ahead and drop '!= 0' for the standard error check pattern

source/credentials_provider_imds.c
```diff
@@ -333,9 +437,27 @@ static void s_imds_on_stream_complete_fn(struct aws_http_stream *stream, int err
...
+            AWS_LOGF_ERROR(AWS_LS_AUTH_CREDENTIALS_PROVIDER, "Failed to process IMDS token response.");
+            s_imds_finalize_get_credentials_query(imds_user_data);
+        }
+        return;
+    }
+
+    // Unauthorized for other requests, means service requires token.
+    if (imds_user_data->status_code == 401) {
```
bretambrose: For status codes that aren't as well known, I would make a constant with both the number and the name in it (ala HTTP_STATUS_CODE_401_BLAH_BLAH)

include/aws/auth/credentials.h
```diff
@@ -155,7 +155,8 @@ struct aws_credentials_provider_chain_options {
 struct aws_credentials_provider_imds_options {
     struct aws_credentials_provider_shutdown_options shutdown_options;
     struct aws_client_bootstrap *bootstrap;
-
+    /* if not set, the default value means token required */
+    bool token_not_required;
```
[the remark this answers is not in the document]
singku-china: The new request flow is described in SEP https://code.amazon.com/packages/AwsDrSeps/blobs/39c7e8e181ee7a9705af1e4670615cd6798c35c2/--/seps/accepted/shared/imds_specifications.md

There are formal names `IMDSV1` and `IMDSV2`
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html
>The command format is different, depending on whether you use IMDSv1 or IMDSv2.

I agree that change it to the formal name would make more sense for users.

source/credentials_provider_imds.c
singku-china: I tried use the original version, but have to do function pointer type conversion for the prediction argument. Then gcc will give warning.

source/credentials_provider_imds.c
singku-china: I wish these were defined in aws-c-http

include/aws/auth/credentials.h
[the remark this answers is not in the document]
singku-china: For the documentation, I think I can move the one for `aws_imds_query_state` to be here.

source/credentials_provider_imds.c
bretambrose: I mean just replace "unsigned char" with "uint8_t", as per the definition of aws_byte_predicate_fn.

source/credentials_provider_imds.c
```diff
@@ -100,22 +114,26 @@ static struct aws_credentials_provider_imds_user_data *s_aws_credentials_provide
...
     wrapped_user_data->imds_provider = imds_provider;
+    aws_credentials_provider_acquire(imds_provider);
     wrapped_user_data->original_user_data = user_data;
     wrapped_user_data->original_callback = callback;
 
     if (aws_byte_buf_init(&wrapped_user_data->current_result, imds_provider->allocator, IMDS_RESPONSE_SIZE_INITIAL)) {
         goto on_error;
     }
 
+    if (aws_byte_buf_init(&wrapped_user_data->token_result, imds_provider->allocator, IMDS_RESPONSE_SIZE_INITIAL)) {
```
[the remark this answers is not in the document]
bretambrose: Anything less than 4 digits is fine.  But I would use a named constant.

justinboswell:
Depends on: https://github.com/awslabs/aws-c-cal/pull/37

include/aws/auth/credentials.h
```diff
@@ -152,11 +152,16 @@ struct aws_credentials_provider_chain_options {
...
+ * Prior to these two requests, IMDSV2 takes one more token (Http PUT) request
+ * to get secure token used in following requests.
+ */
 struct aws_credentials_provider_imds_options {
     struct aws_credentials_provider_shutdown_options shutdown_options;
     struct aws_client_bootstrap *bootstrap;
-    /* if not set, the default value means token required */
-    bool token_not_required;
+    /* If not set, this value will be false, means use IMDSV2 */
+    bool use_imds_v1;
```
bretambrose: Despite there only being two possibilities, would prefer this to be an enum, v1 vs v2.

PR #66, opened 2020-03-16
Updated http calls, to use the new stream activate apis. Rewrote imds creds provider state machinery

JonathanHenson:


source/credentials_provider_sts.c
```diff
@@ -334,16 +336,21 @@ static void s_on_connection_setup_fn(struct aws_http_connection *connection, int
...
-    provider_impl->function_table->aws_http_stream_release(stream);
+        provider_impl->function_table->aws_http_stream_release(stream);
+        return;
+    }
 
-    return;
 error:
+    if (stream) {
+        provider_impl->function_table->aws_http_stream_release(stream);
+    }
```
graebm: it's safe to call release() on a NULL stream, to support naive cleanup in `error:` labels like this

source/credentials_provider_ecs.c
```diff
@@ -460,7 +461,14 @@ static int s_make_ecs_http_query(
     struct aws_http_stream *stream =
         impl->function_table->aws_http_connection_make_request(ecs_user_data->connection, &request_options);
 
-    return stream == NULL ? AWS_OP_ERR : AWS_OP_SUCCESS;
+    if (stream) {
+        if (impl->function_table->aws_http_stream_activate(stream)) {
+            impl->function_table->aws_http_stream_release(stream);
+            goto on_error;
```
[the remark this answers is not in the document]
JonathanHenson: i like yours better.

source/credentials_provider_imds.c
```diff
@@ -565,86 +351,403 @@ static int s_make_imds_http_query(
         .request = request,
     };
 
-    struct aws_http_stream *stream =
-        impl->function_table->aws_http_connection_make_request(imds_user_data->connection, &request_options);
+    imds_user_data->query_state += 1;
```
bretambrose: This works but I can see it blowing up without us knowing due to future change in the enum.  At bare minimum we should document the state enum and every place we modify it arithmetically with warnings/disclaimers about changing the state enum.  Alternatively, we could add advance/rewind functions for the state enum that operate on the specific enum values.

source/credentials_provider_imds.c
```diff
@@ -653,17 +756,50 @@ static void s_imds_on_acquire_connection(struct aws_http_connection *connection,
...
+    struct aws_credentials_provider_imds_impl *impl = imds_user_data->imds_provider->impl;
+
+    struct aws_http_connection *connection = impl->function_table->aws_http_stream_get_connection(stream);
+    impl->function_table->aws_http_stream_release(stream);
+    impl->function_table->aws_http_connection_manager_release_connection(impl->connection_manager, connection);
+
+    /* try again, just drop the state from the response to the request state by subtracting one.
+     * Don't run the state machine in this callback in this case, let the acquire connection callback handle it. */
+    if (error_code == AWS_ERROR_HTTP_CONNECTION_CLOSED) {
+        AWS_LOGF_DEBUG(
```
bretambrose: Glad to see this here (was about to complain about its absence) but I feel like this needs a disclaimer along the lines of "temporary/TODO: add retry strategy hook when that system gets implemented"

source/credentials_provider_imds.c
```diff
@@ -653,17 +756,50 @@ static void s_imds_on_acquire_connection(struct aws_http_connection *connection,
...
+            AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+            "id=%p: Connection was closed, retrying the last request on a new connection.",
+            (void *)imds_user_data->imds_provider);
+        /* roll back to the last request we made, and let it retry. */
+        imds_user_data->query_state -= 1;
```
bretambrose: Per above I'd wrap this in a function.  Even if you keep the inc/dec, you can assert that it's an advanceable/rollbackable state too.

source/credentials_provider_imds.c
```diff
@@ -653,17 +756,50 @@ static void s_imds_on_acquire_connection(struct aws_http_connection *connection,
...
+    } else {
+        s_query_state_machine[imds_user_data->query_state](imds_user_data);
+    }
+
+    /* if there's more work to do, acquire a connection, and run the machine again. */
+    if (imds_user_data->query_state < AWS_IMDS_QS_COMPLETE) {
```
bretambrose: Replace with s_is_terminal_state(...) or something.

source/credentials_provider_imds.c
bretambrose: Actually, is there a not-terrible temporary hack we can do here to make sure we always terminate in a reasonable amount of time even in the face of something pathological (repeated connection closes)?

JonathanHenson: I'll put in a max retry count.

source/credentials_provider_imds.c
JonathanHenson: this was actually a bug waiting to happen. The state functions are responsible for state changes. We just lucked out that I was setting it after the call to an explicit value.

JonathanHenson: deleted

source/credentials_provider_imds.c
```diff
@@ -737,14 +740,29 @@ static imds_state_fn *s_query_state_machine[] = {
...
 };
 
+static inline void s_imds_state_machine_roll_back_to_request_state(
+    struct aws_credentials_provider_imds_user_data *user_data) {
+    AWS_FATAL_ASSERT(
+        user_data->query_state < AWS_IMDS_QS_COMPLETE &&
+        "State machine can't be advanced, it's already in a terminal state.");
+    user_data->query_state -= 1;
+    /* request states are evenly numbered. */
+    AWS_FATAL_ASSERT(!(user_data->query_state & 0x01));
```
bretambrose: If we're adding string context on the assert, might as well indicate something like "cannot rollback from request state"

bretambrose: s_is_request_state, s_is_response_state pretty plz

source/credentials_provider_imds.c
```diff
@@ -782,18 +801,27 @@ static void s_imds_on_stream_complete_fn(struct aws_http_stream *stream, int err
...
+            s_imds_state_machine_roll_back_to_request_state(imds_user_data);
+        } else {
+            AWS_LOGF_ERROR(
+                AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+                "id=%p: Connection was closed, retries have been exhausted.",
+                (void *)imds_user_data->imds_provider);
+            /* roll back to the last request we made, and let it retry. */
+            imds_user_data->query_state = AWS_IMDS_QS_UNRECOVERABLE_ERROR;
+        }
     } else {
```
bretambrose: are there any other error codes that need a check? or termination?

source/credentials_provider_imds.c
```diff
@@ -565,86 +354,418 @@ static int s_make_imds_http_query(
...
+    struct aws_http_header headers[] = {
+        token_ttl_header,
+    };
+
+    struct aws_byte_cursor verb = aws_byte_cursor_from_c_str("PUT");
+
+    imds_user_data->query_state = AWS_IMDS_QS_TOKEN_RESP;
+
+    if (s_make_imds_http_query(imds_user_data, &verb, &uri, headers, AWS_ARRAY_SIZE(headers))) {
+        imds_user_data->query_state = AWS_IMDS_QS_QUERY_NEVER_CLEARED_STACK;
```
justinboswell: Could we set this before the call to s_make_imds_http_query? That way there is 0 chance of racing ever?

source/credentials_provider_imds.c
JonathanHenson: shouldn't be

source/credentials_provider_imds.c
```diff
@@ -565,86 +354,418 @@ static int s_make_imds_http_query(
...
 
-on_error:
+    if (impl->function_table->aws_http_stream_activate(stream)) {
+        goto on_error;
+    }
+
+    return AWS_OP_SUCCESS;
 
+on_error:
+    impl->function_table->aws_http_stream_release(stream);
```
justinboswell: Should we set the error state here on the user data?

source/credentials_provider_imds.c
```diff
@@ -565,86 +354,418 @@ static int s_make_imds_http_query(
...
+    if (imds_user_data->status_code == AWS_HTTP_STATUS_CODE_401_UNAUTHORIZED) {
+        s_aws_credentials_provider_imds_user_data_reset_scratch_data(imds_user_data);
+        imds_user_data->token_required = true;
+        imds_user_data->query_state = AWS_IMDS_QS_TOKEN_REQ;
+        return;
+    }
+
+    /*
+     * At this step, on anything other than a 200, nullify the
+     * response and pretend there was an error
```
justinboswell: Nitpick: Is it really pretending? Isn't it REALLY an error? :)

source/credentials_provider_imds.c
```diff
@@ -565,86 +354,418 @@ static int s_make_imds_http_query(
...
+    struct aws_byte_cursor null_terminator_cursor = aws_byte_cursor_from_string(s_empty_string);
+    if (aws_byte_buf_append_dynamic(&imds_user_data->current_result, &null_terminator_cursor)) {
+        imds_user_data->query_state = AWS_IMDS_QS_UNRECOVERABLE_ERROR;
+        aws_raise_error(AWS_AUTH_PROVIDER_PARSER_UNEXPECTED_RESPONSE);
+        goto done;
+    }
+
+    document_root = cJSON_Parse((const char *)imds_user_data->current_result.buffer);
+    if (document_root == NULL) {
+        AWS_LOGF_ERROR(AWS_LS_AUTH_CREDENTIALS_PROVIDER, "Failed to parse IMDS response as Json document.");
```
justinboswell: Nitpick: JSON or json

source/credentials_provider_imds.c
justinboswell: Discussed this, it's not a problem.

source/credentials_provider_imds.c
JonathanHenson: this isn't a state function. Only state functions are allowed to mutate state.

PR #69, opened 2020-04-15
Fix Queued Async Test.

singku-china:
*Issue #, if available:*
Use this scripts on Mac to reproduce the test failure
```
#!/bin/sh
while [ $? == 0 ]
do
    ./aws-c-auth-tests cached_credentials_provider_queued_async_test
done
```
The root cause are here:
https://github.com/awslabs/aws-c-auth/blob/63ad49d0a75de9ca20f8a235e3cb53aecfd364e1/tests/credentials_provider_utils.c#L316-L318
and

https://github.com/awslabs/aws-c-auth/blob/63ad49d0a75de9ca20f8a235e3cb53aecfd364e1/tests/credentials_provider_utils.c#L237

In a case like this:
L316 wakes up `wait credential callback` in test, then main thread in test heads over to cached_provider release, which will de-re_fcnt the source provider, which is mocked_provider.  Then the worker thread in this function heads over to L318, which releases mocked_provider as well. Then it finds that the ref_cnt is now zero. It will call mock provider destroy.

At L237, within the mocked provider destroy function
The thread which calls `aws_thread_join` is the thread itself, this will raise a `deadlock` error.
The thread will not be joined correctly.

*Description of changes:*
Before calling cached_provider release in test, wait till the ref_cnt of mocked provider becomes one. Then the cached provider can safely release and delete both.

Since it's test, here I just used `check and sleep` to do the `wait till`.


singku-china:
@bretambrose I'll do it after finishing my current process creds task (probably next week). If you want to it before that, by all means. :)

tests/credentials_tests.c
```diff
@@ -571,6 +579,7 @@ static int s_cached_credentials_provider_queued_async_test(struct aws_allocator
         s_verify_callback_status(&callback_results, 2, s_access_key_id_2, s_secret_access_key_2, s_session_token_2) ==
         0);
 
+    s_wait_for_provider_ref_cnt_to_become_one(mock_provider);
```
graebm, line 582: I've got a different idea that gets at the root of the issue.

If the leak is due to the shutdown timing of the background thread used by the mock-async-provider, maybe we should use the `aws_global_thread_creator_shutdown_wait_for()` mechanism:
1) `aws_global_thread_creator_increment()` when the thread is created
2) `aws_global_thread_creator_decrement()` after the thread is joined and cleaned up
3) Tests using the mock-async-provider end with `aws_global_thread_creator_shutdown_wait_for()`

graebm: This is not the issue. The issue is where the final refcount is released from the background thread so the join here fails

graebm (commented):
Totally different solution:
Get rid of the custom background thread that we were having circular ownership issues with. Instead, schedule tasks on an event-loop. This lets us take advantage of all the tooling we have around event-loop-groups

tests/credentials_provider_utils.c
```diff
@@ -263,72 +247,67 @@ static void mock_async_background_thread_function(void *arg) {
...
+    struct aws_credentials_provider_mock_async_impl *impl = provider->impl;
 
-            aws_array_list_clear(&temp_queries);
-        }
-    }
+    struct aws_task *task = aws_mem_calloc(provider->allocator, 1, sizeof(struct aws_task));
+    AWS_FATAL_ASSERT(task);
+    aws_task_init(
+        task,
+        s_async_mock_credentials_provider_fire_callbacks_task,
```
singku, line 295: event loop group will execute this function as soon as possible, so when this `s_async_mock_credentials_provider_fire_callbacks_task` is executed, how do we know the credentials is ready?

graebm: This helper function is only used by the "mock async credentials provider". The constructor for this class takes an array of credentials. Whenever `aws_credentials_provider_mock_async_fire_callbacks()` function is called, it provides the next credential from that array.

singku: The order of the test is like this:
```
s_invoke_get_credentials
   aws_credentials_provider_get_credentials with aws_test_get_credentials_async_callback

s_wait_for_get_credentials_with_mock_async_provider
   aws_thread_current_sleep
   aws_credentials_provider_mock_async_fire_callbacks
      schedule event task with callback: s_async_mock_credentials_provider_fire_callbacks_task
   aws_wait_on_credentials_callback   

s_verify_callback_status
```

Since this mock async creds provider is an async one, after it invokes the get creds call, it won't wait, just return and to schedule the event loop task, which is `s_async_mock_credentials_provider_fire_callbacks_task`, but this task callback will be executed asap, when it's executed, it won't know whether the creds in the previous invoke creds get call is ready or not, will it?

Or is `aws_thread_current_sleep` make sure it will get?

graebm: The mock credentials provider already has credentials (the constructor takes an array of credentials), but won't "provide" them to anyone until `aws_credentials_provider_mock_async_fire_callbacks()` is called.

I added some comment to hopefully make this more clear

singku (approved):

bretambrose (approved):

PR #71, opened 2020-04-22
Support Process Credentials Provider

singku-china:
*Issue #, if available:*

*Description of changes:*
This PR is based on the GetCmdOutput PR of aws-c-common.


include/aws/auth/credentials.h
```diff
@@ -266,6 +266,32 @@ struct aws_credentials_provider_sts_options {
...
+ * Note that our default provider chain will not add this provider, it's SDKs' choice to add this provider
+ * to their provider chain.
+ */
+struct aws_credentials_provider_process_options {
+    struct aws_credentials_provider_shutdown_options shutdown_options;
+    /**
+     * In which profile name to look for credential_process,
+     * if not provided, we will try environment variable: AWS_PROFILE.
+     */
+    struct aws_string *profile_to_use;
```
bretambrose, line 292: byte cursor less restrictive than a string

include/aws/auth/credentials.h
```diff
@@ -266,6 +266,32 @@ struct aws_credentials_provider_sts_options {
     aws_io_clock_fn *system_clock_fn;
 };
 
+/**
+ * The process credentials provider sources credentials from running a command or process.
+ * The command is configured in config file with the config key as "credential_process".
```
bretambrose, line 271: I'd reword this along the lines of:

"The command to run is sourced from a profile in the AWS config file, using the standard profile selection rules.  The profile key the command is read from is "credential_process." 

And then maybe give a very simple example ala

[default]
credential_process=/usr/local/bin/my-credential-fetcher --stage=Prod

include/aws/auth/credentials.h
```diff
@@ -266,6 +266,32 @@ struct aws_credentials_provider_sts_options {
...
+ * On successfully running the command, the output should be a json data with the following
+ * format:
+ * {
+    "Version": 1,
+    "AccessKeyId": "accesskey",
+    "SecretAccessKey": "secretAccessKey"
+    "SessionToken": "....",
+    "Expiration": "2019-05-29T00:21:43Z"
+   }
+ * Version here identify the command output format version. If it involved in future, SDK should update
```
bretambrose, line 281: "identifies" in first sentence.  Remove the rest of the paragraph and replace with "This provider is not a part of the default provider chain."  I know I do it sometimes, but I feel like "If X happens in the future then we should do Y" notes to ourselves should stay out of the public API documentation.

source/credentials_provider_process.c
```diff
@@ -0,0 +1,374 @@
...
+static struct aws_credentials *s_parse_credentials_from_json_document(
+    struct aws_allocator *allocator,
+    struct aws_byte_buf *document) {
+
+    struct aws_credentials *credentials = NULL;
+    cJSON *document_root = NULL;
+    bool success = false;
+    bool parse_error = true;
+
+    document_root = cJSON_Parse((const char *)document->buffer);
```
bretambrose, line 48: Do the keys here match any of the other JSON doc providers? Any way to refactor such that we only do this once?  I feel like I've seen this function 2-3 times.  Maybe something that takes a JSON doc, the key of the primary section where everything is and then the 3 keys used to pull credential ccomponents?

source/credentials_provider_process.c
```diff
@@ -0,0 +1,374 @@
...
+
+    /*
+     * Build the credentials
+     */
+    struct aws_byte_cursor access_key_id_cursor = aws_byte_cursor_from_c_str(access_key_id->valuestring);
+    struct aws_byte_cursor secret_access_key_cursor = aws_byte_cursor_from_c_str(secret_access_key->valuestring);
+    struct aws_byte_cursor session_token_cursor = aws_byte_cursor_from_c_str(session_token->valuestring);
+    struct aws_byte_cursor creds_expiration_cursor = aws_byte_cursor_from_c_str(creds_expiration->valuestring);
+
+    if (access_key_id_cursor.len == 0 || secret_access_key_cursor.len == 0 || session_token_cursor.len == 0) {
```
bretambrose, line 94: The SEP requires session token?  That's unexpected.

source/credentials_provider_process.c
```diff
@@ -0,0 +1,374 @@
...
+}
+
+static int s_get_credentials_from_process(
+    struct aws_credentials_provider *provider,
+    aws_on_get_credentials_callback_fn callback,
+    void *user_data) {
+
+    struct aws_credentials_provider_process_impl *impl = provider->impl;
+    struct aws_credentials *credentials = NULL;
+    struct aws_run_command_options options = {.command = aws_string_c_str(impl->command)};
```
bretambrose, line 146: put a comma after the closing paranthese to get the nice multi-line formatting

come to think of it, the command member should have been a byte cursor

source/credentials_provider_process.c
```diff
@@ -0,0 +1,374 @@
...
+    } else {
+        AWS_LOGF_ERROR(
+            AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+            "Failed to build config profile collection from file at (%s) : %s",
+            aws_string_c_str(config_file_path),
+            aws_error_str(aws_last_error()));
+        goto on_error;
+    }
+
+    aws_string_destroy(config_file_path);
```
bretambrose, line 228: Seems like this could all be combined and on_error: become done:

source/credentials_provider_process.c
```diff
@@ -0,0 +1,374 @@
...
+    AWS_ZERO_STRUCT(*provider);
+    AWS_ZERO_STRUCT(*impl);
+
+    command = s_get_command(allocator, options->profile_to_use);
+    if (!command) {
+        goto on_error;
+    }
+
+    struct aws_byte_cursor command_cursor = aws_byte_cursor_from_string(command);
+    if (aws_byte_buf_init_copy_from_cursor(&command_with_stderr_redirected, allocator, command_cursor)) {
```
bretambrose, line 344: maybe fold this into s_get_command

source/credentials_provider_process.c
```diff
@@ -0,0 +1,374 @@
...
+            aws_string_destroy(*target);
+        }
+        struct aws_profile_property *property = aws_profile_get_property(profile, config_key);
+        if (property) {
+            *target = aws_string_new_from_string(allocator, property->value);
+        }
+    }
+}
+
+static struct aws_string *s_get_command(struct aws_allocator *allocator, const struct aws_string *profile_to_use) {
```
bretambrose, line 257: Looking at this makes me realize that I messed up the profile api.  Some of these params should be byte cursors, and arguably byte_buf output params or cursor out params.  

We only use it internally so I am tempted to rework it maybe this weekend.  I think it would simplify your usage quite a bit.  A single byte buf in the implementation to hold the command, pulling from the profile would append into it, then you'd append the std err redirection and then you'd pass a cursor of the buf to the aws_run_process.  Most of the temporaries and their cleanup (in multiple spots) would go away.

source/credentials_provider_process.c
```diff
@@ -0,0 +1,374 @@
...
+
+    bool success = false;
+    struct aws_string *command = NULL;
+    struct aws_profile_collection *config_profile = NULL;
+    struct aws_string *profile_name = NULL;
+    struct aws_profile *profile = NULL;
+
+    config_profile = s_load_profile(allocator);
```
bretambrose, line 265: this is a collection not an individual profile

source/credentials_provider_process.c
```diff
@@ -0,0 +1,374 @@
...
+        profile = aws_profile_collection_get_profile(config_profile, profile_name);
+    }
+
+    if (!profile) {
+        AWS_LOGF_ERROR(
+            AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+            "Failed to resolve config profile during process credentials provider initialization.");
+        goto on_finish;
+
+    } else {
```
bretambrose, line 281: else isn't necessary here

source/credentials_provider_process.c
```diff
@@ -0,0 +1,374 @@
...
+    aws_string_destroy(config_file_path);
+    aws_profile_collection_destroy(config_profiles);
+    return NULL;
+}
+
+static void s_check_or_get_with_profile_config(
+    struct aws_allocator *allocator,
+    struct aws_profile *profile,
+    struct aws_string **target,
+    const struct aws_string *config_key) {
```
bretambrose, line 241: I'd make the output param the last param

source/credentials_provider_process.c
```diff
@@ -0,0 +1,374 @@
...
+
+    if (!allocator || !profile || !config_key) {
```
bretambrose, line 243: I don't think there's a need to allocate anything here or clean up an existing value.

source/credentials_provider_process.c
```diff
@@ -0,0 +1,374 @@
...
+        sizeof(struct aws_credentials_provider_process_impl));
+
+    if (!provider) {
+        goto on_error;
+    }
+
+    AWS_ZERO_STRUCT(*provider);
+    AWS_ZERO_STRUCT(*impl);
+
+    command = s_get_command(allocator, options->profile_to_use);
```
bretambrose, line 338: I would change command to a byte buf and have s_get_command write into it.  Then fold the stderr redirection into s_get_command by just doing an append to the impl's byte buf.

source/credentials_provider_process.c
singku-china: Ok, then I will address these together after your rework.

source/credentials_provider_process.c
singku-china: If *target is not null but the length is zero, I think still need to destroy and allocate new one. change it to byte_buf would be easier. The part of byte_buf I don't like is the complexity to process trimmed data.

source/credentials_provider_process.c
singku-china: I can rework this function to be a shared one.

source/credentials_provider_process.c
singku-china: the command member of options? or impl? If it's impl, then I disagree, we need to make sure the memory store command won't disappear unexpectedly.

source/credentials_provider_process.c
singku-china: you are right, it's not required.

include/aws/auth/private/credentials_utils.h
```diff
@@ -109,6 +108,61 @@ void aws_credentials_provider_destroy(struct aws_credentials_provider *provider)
 AWS_AUTH_API
 void aws_credentials_provider_invoke_shutdown_callback(struct aws_credentials_provider *provider);
 
+struct aws_parse_credentials_from_json_doc_options {
+    char *access_key_id_name;
```
bretambrose, line 112: const on the char *s

include/aws/auth/private/credentials_utils.h
```diff
@@ -109,6 +108,61 @@ void aws_credentials_provider_destroy(struct aws_credentials_provider *provider)
...
+ * except it accpets an char buffer json document as it's input.
+ */
+AWS_AUTH_API
+struct aws_credentials *aws_parse_credentials_from_json_document(
+    struct aws_allocator *allocator,
+    const char *json_document,
+    const struct aws_parse_credentials_from_json_doc_options *options);
+
+AWS_AUTH_API
+int aws_append_null_terminator_to_byte_buf(struct aws_byte_buf *buf);
```
bretambrose, line 164: I'd move to common and rename as aws_byte_buf_append_null_terminator or ..._zero_byte

PR #72, opened 2020-04-28
Fix Address Sanitizer reported bug

singku-china:
*Issue #, if available:*

*Description of changes:*


source/credentials_provider_ecs.c
```diff
@@ -545,11 +545,10 @@ static void s_credentials_provider_ecs_destroy(struct aws_credentials_provider *
         return;
     }
 
-    impl->function_table->aws_http_connection_manager_release(impl->connection_manager);
-
     aws_string_destroy(impl->path_and_query);
     aws_string_destroy(impl->auth_token);
     aws_tls_connection_options_clean_up(&impl->connection_options);
+    impl->function_table->aws_http_connection_manager_release(impl->connection_manager);
```
singku-china, line 551: Release will call on_connection_manager_shutdown, which will do memory free for provider and impl. We should free impl related first, then call release. Same to x509, sts_web_identity.

source/credentials_provider_imds.c
```diff
@@ -102,10 +103,12 @@ struct aws_credentials_provider_imds_user_data {
     bool token_required;
     uint8_t retry_count;
     int status_code;
+
+    struct aws_atomic_var ref_count;
+    bool callback_invoked;
```
singku-china, line 108: This is to prevent callback from being invoked multiple times, currently there are two states related to the same state function: s_query_error

tests/credentials_provider_sts_web_identity_tests.c
```diff
@@ -462,10 +461,11 @@ static int s_credentials_provider_sts_web_identity_new_destroy_from_config(struc
     s_aws_wait_for_provider_shutdown_callback();
 
     /* Because we mock the http connection manager, we never get a callback back from it */
-    aws_mem_release(provider->allocator, provider);
     struct aws_credentials_provider_sts_web_identity_impl *impl = provider->impl;
     aws_tls_ctx_destroy(impl->ctx);
     aws_tls_connection_options_clean_up(&impl->connection_options);
+    aws_mem_release(provider->allocator, provider);
```
singku-china, line 467: provider and impl's memory are allocated together.

source/credentials_provider_ecs.c
justinboswell: Please add this as a comment here. This is important to the next reader.

source/credentials_provider_imds.c
```diff
@@ -709,12 +729,15 @@ static void s_imds_query_complete(struct aws_credentials_provider_imds_user_data
...
 }
 
-static void s_imds_query_error(struct aws_credentials_provider_imds_user_data *imds_user_data) {
+static void s_imds_query_error(struct imds_user_data *imds_user_data) {
     AWS_PRECONDITION(
         imds_user_data->query_state == AWS_IMDS_QS_UNRECOVERABLE_ERROR ||
         imds_user_data->query_state == AWS_IMDS_QS_QUERY_NEVER_CLEARED_STACK);
 
-    imds_user_data->original_callback(NULL, imds_user_data->original_user_data);
+    if (!imds_user_data->callback_invoked) {
```
justinboswell, line 737: What is making this bool thread safe? is there a lock held above this callback?

source/credentials_provider_sts_web_identity.c
```diff
@@ -690,11 +690,10 @@ static void s_credentials_provider_sts_web_identity_destroy(struct aws_credentia
         return;
     }
 
-    impl->function_table->aws_http_connection_manager_release(impl->connection_manager);
-
     aws_string_destroy(impl->role_arn);
     aws_string_destroy(impl->role_session_name);
     aws_string_destroy(impl->token_file_path);
+    impl->function_table->aws_http_connection_manager_release(impl->connection_manager);
```
justinboswell, line 696: Add that same comment here too about why this is done this way, and where the final teardown is done

source/credentials_provider_imds.c
singku-china: `imds_user_data` is associated to single credentials request. The workflow is processed by a single thread.

JonathanHenson: it's not all in the same thread on the same connection. It is guaranteed to be sequentially accessed though, and any processor with cache coherency policies (all of them) are fine.

singku-china:
Merged that PR over to here.

source/credentials_provider_imds.c
```diff
@@ -779,18 +806,41 @@ static void s_imds_on_acquire_connection(struct aws_http_connection *connection,
 
         imds_user_data->query_state = AWS_IMDS_QS_QUERY_NEVER_CLEARED_STACK;
     } else {
+        // prevent user_data from being destroyed under the hood.
```
JonathanHenson, line 809: c++ comment

PR #79, opened 2020-05-22
Move STS With Web Identity Creds Provider's parameter from query to H…

singku-china:
…ttp body.

*Issue #, if available:*

*Description of changes:*


source/credentials_provider_sts_web_identity.c
```diff
@@ -95,12 +106,8 @@ static void s_user_data_destroy(struct sts_web_identity_user_data *user_data) {
         impl->function_table->aws_http_connection_manager_release_connection(
             impl->connection_manager, user_data->connection);
     }
-
+    s_user_data_reset_request_and_response(user_data);
```
singku-china, line 109: reset won't clean up buf, just reset len and ptr, so as to avoid buf alloc/dealloc multiple times. Then we need clean up the buf during destroy

singku-china:
Couple of more changes:
- Added more headers: `content-length` `content-type header` `host` `x-amz-api-version`
- Changed http method from `Get` to `Post`.
- The payload buffer shouldn't be destroyed until the request is done.
- For failure test, we should set up an empty config file, otherwise, if the default config file has all we need to initiate a provider, the test would fail.

source/credentials_provider_sts_web_identity.c
```diff
@@ -564,102 +607,98 @@ static int s_make_sts_web_identity_http_query(
...
+    struct aws_credentials_provider_sts_web_identity_impl *impl = user_data->sts_web_identity_provider->impl;
 
     /* "Clear" the result */
-    s_user_data_reset_response(sts_web_identity_user_data);
+    s_user_data_reset_request_and_response(user_data);
 
     /*
-     * Calculate query string:
-     * "/?Action=AssumeRoleWithWebIdentity"
+     * Calculate body message:
```
bretambrose, line 625: Is there documentation or a link on this data and its format?

singku-china: https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html
There is an example at the bottom of the page, but the data is on query uri, also some of the fields are not required.

source/credentials_provider_sts_web_identity.c
```diff
@@ -529,28 +520,35 @@ static int s_make_sts_web_identity_http_query(
         goto on_error;
     }
 
-    if (aws_http_message_set_request_path(request, *uri)) {
+    input_stream = aws_input_stream_new_from_cursor(user_data->allocator, body_cursor);
+    if (!input_stream) {
+        goto on_error;
+    }
+    user_data->input_stream = input_stream;
```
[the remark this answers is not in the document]
singku-china: Figured that input_stream is accessible from request. So I will delete the reference in user_data.

source/credentials_provider_sts_web_identity.c
bretambrose: Can we add a documentation link here then?

source/credentials_provider_sts_web_identity.c
```diff
@@ -470,87 +470,129 @@ static void s_on_stream_complete_fn(struct aws_http_stream *stream, int error_co
...
         .on_response_body = s_on_incoming_body_fn,
         .on_complete = s_on_stream_complete_fn,
-        .user_data = sts_web_identity_user_data,
+        .user_data = user_data,
         .request = request,
     };
 
-    stream = impl->function_table->aws_http_connection_make_request(
-        sts_web_identity_user_data->connection, &request_options);
+    stream = impl->function_table->aws_http_connection_make_request(user_data->connection, &request_options);
```
bretambrose, line 595: We still have the problem where any error from here til the end leaves us with user_data holding a destroyed request.

singku-china: You are damn right. There are similar problems in other providers which trigger http request. I have changed them as well.

singku-china:
Agreed and updated.

PR #81, opened 2020-05-28
Basic IMDSV2 EC2 Metadata Client

singku-china:
*Issue #, if available:*

*Description of changes:*
Added EC2 IMDSV1/V2 based MetadataClient and tests. Will added more wrapper APIs in new PR once this base client is approved. Those wrapper APIs will be used to acquire specific metadata information. Also when wrapper APIs are done. IMDS Creds Provider will be refactored as well.


singku-china:
Tested on EC2 that the client can be used to access meta-data info correctly.

source/aws_imds_client.c
```diff
@@ -0,0 +1,812 @@
...
+    .aws_http_stream_get_incoming_response_status = aws_http_stream_get_incoming_response_status,
+    .aws_http_stream_release = aws_http_stream_release,
+    .aws_http_connection_close = aws_http_connection_close,
+};
+
+static void s_aws_imds_client_destroy(struct aws_imds_client *client) {
+    if (!client) {
+        return;
+    }
+    aws_retry_strategy_release(client->retry_strategy);
```
bretambrose, line 66: I feel like it's worth documenting that this function will only get called after all in-flight requests are finished and thus nothing is going to try and access the retry strategy after this point.

source/aws_imds_client.c
```diff
@@ -0,0 +1,812 @@
...
+
+    struct aws_http_connection *connection = client->function_table->aws_http_stream_get_connection(stream);
+    client->function_table->aws_http_stream_release(stream);
+    client->function_table->aws_http_connection_manager_release_connection(client->connection_manager, connection);
+
+    /* try again, just drop the state from the response to the request state by subtracting one.
+     * Don't run the state machine in this callback in this case, let the acquire connection callback handle it.
+     * Note these are connection level errors, not http level. Since we obviously connected, it's likely
+     * we're on EC2, plus we have max retries so it's likely safer to just retry everything.*/
+    if (error_code) {
```
bretambrose, line 724: Should log the error here if it's not logged elsewhere.

source/aws_imds_client.c
```diff
@@ -0,0 +1,812 @@
...
+    struct aws_imds_client_user_data *imds_user_data = user_data;
+
+    if (!error_code) {
+        struct aws_imds_client *client = imds_user_data->client;
+        imds_user_data->retry_token = token;
+
+        client->function_table->aws_http_connection_manager_acquire_connection(
+            client->connection_manager, s_on_acquire_connection, imds_user_data);
+    } else {
+        imds_user_data->error_code = error_code;
```
bretambrose, line 784: possibly log success and failure here?

source/aws_imds_client.c
```diff
@@ -0,0 +1,812 @@
...
+#include <ctype.h>
+
+#if defined(_MSC_VER)
+#    pragma warning(disable : 4204)
+#endif /* _MSC_VER */
+
+/* instance role credentials body response is currently ~ 1300 characters + name length */
+#define IMDS_RESPONSE_SIZE_INITIAL 2048
+#define IMDS_RESPONSE_TOKEN_SIZE_INITIAL 64
+#define IMDS_RESPONSE_SIZE_LIMIT 10000
```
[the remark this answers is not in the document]
singku-china: Didn't find a document about this. So far the largest response I have seen is credential, which is around 1K. Do you think we should set a limit on this?

source/aws_imds_client.c
```diff
@@ -0,0 +1,812 @@
...
+    socket_options.connect_timeout_ms = (uint32_t)aws_timestamp_convert(
+        IMDS_CONNECT_TIMEOUT_DEFAULT_IN_SECONDS, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_MILLIS, NULL);
+
+    struct aws_http_connection_manager_options manager_options;
+    AWS_ZERO_STRUCT(manager_options);
+    manager_options.bootstrap = options->bootstrap;
+    manager_options.initial_window_size = IMDS_RESPONSE_SIZE_LIMIT;
+    manager_options.socket_options = &socket_options;
+    manager_options.tls_connection_options = NULL;
+    manager_options.host = aws_byte_cursor_from_c_str("169.254.169.254");
```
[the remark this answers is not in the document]
singku-china: So do you recommend `host` can be configured via options? I believe even when ipv6 is supported, the address should be a fixed one?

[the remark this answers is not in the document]
bretambrose: I'm not entirely sure.  There may be a DNS option such that it looks up and uses one automatically based on the underlying instance configuration.  I don't think you need to do anything, but do think about how you might expose in the future.

source/aws_imds_client.c
```diff
@@ -460,11 +587,28 @@ static int s_make_imds_http_query(
...
 /* Make an http request to put a ttl and hopefully get a token back. */
 static void s_query_token(struct aws_imds_client_user_data *user_data) {
     AWS_PRECONDITION(user_data->query_state == AWS_IMDS_QS_TOKEN_REQ);
 
+    /**
+     * If this requester shouldn't update token, meaning either the token is available,
+     * or the token is updating by other requester, this requester should just either copy
+     * token or wait for the update then copy.
+     */
+    while (!s_requester_test_and_try_dominate_cached_token_update(user_data)) {
```
bretambrose, line 601: Is this spinning on a pending token query?  If so, we'll need to rework this approach substantially.  The cached credentials provider faces a similar situation: arbitrary incoming async get requests but the "real" singleton query could be outstanding or need to be made.  It solves it by actually tracking all the requests and completing them when the query completes either successfully or unsuccessfully.  Meanwhile those get requests return immediately; we can't block or spin anywhere.

source/aws_imds_client.c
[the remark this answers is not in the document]
singku-china: Since currently the mechanism is unknown, let's keep it this way for now, and make changes later based on a more clear mechanism.

source/aws_imds_client.c
```diff
@@ -955,6 +873,197 @@ static void s_on_retry_token_acquired(
...
+            struct aws_byte_cursor cursor = aws_byte_cursor_from_buf(&client->cached_token);
+            if (aws_byte_buf_append_dynamic(&user_data->token_result, &cursor)) {
+                ret = AWS_IMDS_TK_CP_RS_UNEXPECTED_ERROR;
+            } else {
+                ret = AWS_IMDS_TK_CP_RS_SUCCESS;
+            }
+            break;
+        }
+
+        case AWS_IMDS_TK_ST_INVALID: {
```
bretambrose, line 894: I feel like there's some simplifications here that can be made:

(1) The validity of the client's token feels like a client property and so refreshing and updating  (and retrying) it should be coupled to the client itself.  Can token queries not be associated with a request?  In particular, everything could go in the waiting queue and the user data on the token query could be the client itself.  This would let you collapse UPDATING and WAITING_IN_QUEUE states, everything is now WAITING_IN_QUEUE.  

(2) Once token refreshes are uncoupled from an individual request, you could also simplify the logic of s_imds_client_update_token_safely:

list temp_list;

lock
if success
  swap(queue, temp_list)
else
  if retry_token_failure
    swap(queue, temp_list)
   
unlock

iterate(temp_list)
  completion callback(success or failure)

The loop is now out of the lock, and retry is all or nothing since it's now associated with the client.  The loop's logic is simpler since all you do is the completion callback.

singku-china: It might be good to move token request to client itself to uncouple the logic in requester. Here I did it in requester is to make sure even though one token query failed (tmp network issue), other requesters in the queue still get the chance to acquire token.

singku-china: I have changed the token logic to client and removed state machine

source/aws_imds_client.c
```diff
@@ -0,0 +1,953 @@
...
+            "id=%p: Connection was closed with error code %d(%s)",
+            (void *)client,
+            error_code,
+            aws_error_str(error_code));
+
+        if (!aws_retry_strategy_schedule_retry(
+                imds_user_data->retry_token, AWS_RETRY_ERROR_TYPE_TRANSIENT, s_on_retry_ready, user_data)) {
+            AWS_LOGF_DEBUG(
+                AWS_LS_IMDS_CLIENT,
+                "id=%p: Connection was closed, retrying the last request on a new connection.",
```
bretambrose, line 690: "Stream complete"

source/aws_imds_client.c
```diff
@@ -0,0 +1,953 @@
...
+    int error_code,
+    struct aws_retry_token *token,
+    void *user_data) {
+    (void)strategy;
+
+    struct imds_user_data *imds_user_data = user_data;
+    struct aws_imds_client *client = imds_user_data->client;
+
+    if (!error_code) {
+        AWS_LOGF_WARN(AWS_LS_IMDS_CLIENT, "id=%p: IMDS Client successfully acquired retry token.", (void *)client);
```
bretambrose, line 720: I'd make this debug or info

source/aws_imds_client.c
```diff
@@ -0,0 +1,953 @@
...
+                ret = AWS_IMDS_TCR_UNEXPECTED_ERROR;
+                aws_linked_list_swap_contents(&pending_queries, &client->pending_queries);
+            } else {
+                client->token_state = AWS_IMDS_TS_UPDATE_IN_PROGRESS;
+            }
+        }
+    }
+    aws_mutex_unlock(&client->token_lock);
+
+    /* poll swapped out pending queries if there is any */
```
bretambrose, line 773: Feels like completing queries could be a separate function that takes the list, error, token copy

source/aws_imds_client.c
```diff
@@ -0,0 +1,953 @@
...
+ */
+static bool s_update_token_safely(struct aws_imds_client *client, struct aws_byte_buf *token, bool token_required) {
+    AWS_FATAL_ASSERT(client);
+    bool updated = false;
+
+    struct aws_linked_list pending_queries;
+    aws_linked_list_init(&pending_queries);
+
+    aws_mutex_lock(&client->token_lock);
+    client->token_required = token_required;
```
bretambrose, line 845: This feels weird, but I guess it's more of a 3-value state: Unknown, Required, Not required?  I don't know if I'd change anything, it's just odd.

source/aws_imds_client.c
singku-china: makes sense.

source/aws_imds_client.c
singku-china: It's either V2 required or V1 not required.  He can set either one, client will find it out for him. A more clear setting would reduce unnecessary network request. Maybe we should document the enum better?

source/aws_imds_client.c
```diff
@@ -0,0 +1,953 @@
...
+            user_data->error_code = AWS_ERROR_UNKNOWN;
+        }
+        s_query_complete(user_data);
+    }
+}
+
+int s_get_resource_async_with_imds_token(struct imds_user_data *user_data);
+
+static void s_query_complete(struct imds_user_data *user_data) {
+    if (user_data->is_imds_token_request) {
```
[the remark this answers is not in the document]
singku-china: Not quit sure if I got your point. but there is debug log in each following up function.

PR #83, opened 2020-06-03
Full IMDSv2 APIs within EC2 Metadata Client

singku-china:
*Issue #, if available:*

*Description of changes:*
Will add test later

### DIFF
--- a/source/aws_imds_client.c
+++ b/source/aws_imds_client.c
@@ -951,3 +951,598 @@ error:
 
     return AWS_OP_ERR;
 }
+
+/**
+ * Higher level API definitions to get specific IMDS info
+ * Reference:
+ * https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html
+ * https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/util/EC2MetadataUtils.html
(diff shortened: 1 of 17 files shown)


source/aws_imds_client.c
```diff
@@ -951,3 +951,664 @@ int aws_imds_client_get_resource_async(
...
+
+    if (aws_byte_buf_init_copy(&json_data, wrapped_user_data->allocator, resource)) {
+        goto on_finish;
+    }
+
+    if (aws_byte_buf_append_null_terminator(&json_data)) {
+        goto on_finish;
+    }
+
+    struct aws_parse_credentials_from_json_doc_options parse_options = {
```
bretambrose, line 1039: I wonder if there's a generalization here for a parse function that takes the json doc and an array of mappings where a mapping is either a (json path -> pointer to a byte cursor in the target struct) or (json path -> pointer to an array list (of byte cursors) in the target struct) or (json path -> pointer to date time) that would let a single function do the parsing with just different array configurations to handle the different docs/substructures.  Not required just something to consider; you'd lose some of the caching (refollowing the same path repeatedly) unless you added additional complexity.

Also, I imagine this data changes very slowly from a modeling perspective, but it's a little uneasy to have the model hard-coded.  But if you don't provide that modeling then there's nothing to do really.  I'd check with henso to make sure this seems reasonable.  I don't know what other ec2 metadata clients do.

singku-china: Justin mentioned in the stand up that this would be our next year target to do data serialization/deserialization.

PR #86, opened 2020-06-07
Flags + decouple iot ws handshake behavior from header skip callback

bretambrose:
* Move signing config bools into a flags bitfield
* Decouple header skipping from not adding the session token to the canonical request of an iot websocket handshake.

### DIFF
--- a/include/aws/auth/signing_config.h
+++ b/include/aws/auth/signing_config.h
@@ -23,7 +23,7 @@
 
 struct aws_credentials;
 
-typedef bool(aws_should_sign_param_fn)(const struct aws_byte_cursor *name, void *userdata);
+typedef bool(aws_should_sign_header_fn)(const struct aws_byte_cursor *name, void *userdata);
 
 /**
  * A primitive RTTI indicator for signing configuration structs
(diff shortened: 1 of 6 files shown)


include/aws/auth/signing_config.h
```diff
@@ -163,28 +163,36 @@ struct aws_signing_config_aws {
...
+    aws_should_sign_header_fn *should_sign_header;
+    void *should_sign_header_ud;
+
+    struct {
+        /**
+         * We assume the uri will be encoded once in preparation for transmission.  Certain services
+         * do not decode before checking signature, requiring us to actually double-encode the uri in the canonical
+         * request in order to pass a signature check.
+         */
+        uint32_t use_double_uri_encode : 1;
```
graebm, line 183: not sure it's worth the effort here, changing the types and moving them in a sub-struct.
You could keep the external API unchanged and get the same size-saving effect if you just modified them like:
`bool use_double_uri_encode : 1;`

include/aws/auth/signing_config.h
```diff
@@ -163,28 +163,36 @@ struct aws_signing_config_aws {
...
+        /**
+         * Controls whether or not the uri paths should be normalized when building the canonical request
+         */
+        uint32_t should_normalize_uri_path : 1;
+
+        /**
+         * Should the sesssion token query param be omitted from the canonical request?  This is a special case that
+         * should only be true when performing a websocket handshake to succcessfully authenticate to iot core.
+         */
+        uint32_t omit_session_token_query_param : 1;
```
graebm, line 194: comment suggestion:
```
    /**
     * Should the "X-Amz-Security-Token" query param be omitted?
     * Normally, this param is added during query param signing if the credentials have a session token.
     * The only known case where this should be true is when signing a websocket handshake to IoT Core.
     */
```

not sure if variable name should have "security_token" or "session_token"
you're more used to the inconsistent naming of stuff in various places, which is better?

PR #89, opened 2020-06-17
Fix S3 query param signing

graebm:
if not double-encoding URI, don't double-encode query params either

### DIFF
--- a/source/aws_signing.c
+++ b/source/aws_signing.c
@@ -594,11 +594,14 @@ int s_canonical_query_param_comparator(const void *lhs, const void *rhs) {
     const struct aws_uri_param *left_param = lhs;
     const struct aws_uri_param *right_param = rhs;
 
+    /* Note: Sigv4 spec says to sort parameter names by character code point,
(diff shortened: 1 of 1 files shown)


source/aws_signing.c
```diff
@@ -636,17 +639,33 @@ int s_canonical_header_comparator(const void *lhs, const void *rhs) {
...
 
-static int s_append_canonical_query_param(struct aws_uri_param *param, struct aws_byte_buf *buffer) {
-    if (aws_byte_buf_append_encoding_uri_param(buffer, &param->key)) {
-        return AWS_OP_ERR;
+static int s_append_canonical_query_param(
+    struct aws_uri_param *param,
+    struct aws_byte_buf *buffer,
+    bool use_double_uri_encode) {
+
+    if (use_double_uri_encode) {
```
bretambrose, line 647: Maybe add a helper function that does this if else statement?

graebm:
This is not correct either
working on it....

PR #97, opened 2020-08-07
Allow users to pass signed_body_value as string instead of enum.

graebm:
This allows a precalculated hash to be used.

It seems simpler to remove the `aws_signed_body_value_type` enum than keep it around, in addition to a new string.

### BREAKING API WARNING
Downstream CRTs need to be merged immediately after this goes in:
https://github.com/awslabs/aws-crt-python/pull/172
https://github.com/awslabs/aws-crt-cpp/pull/161
https://github.com/awslabs/aws-crt-java/pull/221
https://github.com/awslabs/aws-crt-nodejs/pull/123


source/sigv4_http_request.c
```diff
@@ -294,180 +294,6 @@ struct aws_signable *aws_signable_new_http_request(struct aws_allocator *allocat
     return NULL;
 }
 
-/*
- * Utility struct/API to let us wait on credentials resolution
- */
-struct aws_signing_waiter {
```
graebm, line 300: Deleting this big chunk of code because it wasn't being used by anything

tests/credentials_provider_x509_tests.c
```diff
@@ -584,6 +585,29 @@ static int s_credentials_provider_x509_success_multi_part_doc(struct aws_allocat
 
 AWS_TEST_CASE(credentials_provider_x509_success_multi_part_doc, s_credentials_provider_x509_success_multi_part_doc);
 
+static void s_on_bootstrap_shutdown_complete(void *user_data) {
+    (void)user_data;
```
graebm, line 589: Fixing a crash in an unrelated test that had shutdown timing issues

source/sigv4_http_request.c
bretambrose: It used to be used by elasticurl via dynamic load/bind, but that's a terrible solution and I wish I'd never done it.

PR #99, opened 2020-08-12
signed_body_value as string. fix canonical query param URI-encoding.

graebm:
This PR incorporates 2 major changes that were reviewed separately, but are being merged simultaneously:
1) fix bug where canonical query params were being double URI-encoded. (When this PR was reviewed, it contained only this 1 fix. The details are below...)
2) `signed_body_value` is now a string instead of an enum, allowing users to pass precalculated values. (Originally reviewed here: https://github.com/awslabs/aws-c-auth/pull/97)

---------------------------------------------

### Issue:
aws-c-auth was double URI-encoding the canonical query params.
aws-c-auth extracts the query params from the URI, so it should assume they've already been URI-encoded.

### Solution:
After extracting query params from the URI, decode them. While unencoded, validate the existing query params and add auth query params. Finally, re-encode them all (in the canonical Sigv4 way) when writing to the canonical request.

### Notes:
We COULD have just kept the the query params encoded the whole time, but un-encoding and re-encoding gets around edge cases that could occur if the user encoded the query params with different unreserved characters than Sigv4.


source/aws_signing.c
```diff
@@ -802,6 +830,57 @@ static int s_validate_query_params(struct aws_array_list *params) {
...
 
+/**
+ * Apply or remove URI-encoding to each aws_uri_param in a list.
+ * (new strings are added to temp_strings)
+ */
+static int s_transform_query_params(
+    struct aws_signing_state_aws *state,
+    struct aws_array_list *param_list,
+    struct aws_array_list *temp_strings,
+    int (*byte_buf_append_param_fn)(struct aws_byte_buf *, const struct aws_byte_cursor *)) {
```
bretambrose, line 841: Might add a note that the append function should be dynamic and not fail if there's not enough room

source/aws_signing.c
```diff
@@ -811,30 +890,51 @@ static int s_append_canonical_query_string(struct aws_uri *uri, struct aws_signi
 
     int result = AWS_OP_ERR;
 
+    /* aws-c-auth takes query params from the URI, so they should already be URI-encoded.
+     * To ensure that the signature uses "canonical" URI-encoding, we decode and then re-encode the params. */
+    struct aws_array_list temp_strings;
+    if (aws_array_list_init_dynamic(
+            &temp_strings, state->allocator, INITIAL_QUERY_FRAGMENT_COUNT * 4, sizeof(struct aws_string *))) {
```
bretambrose, line 897: Feels like this could be more precisely calculated if done after the param split below

mullermp:
I tried implementing the "get-vanilla-query-order-encoded" test case in Ruby but I ran into complications, which led me to this PR. Does the CRT SigV4 signer expect URIs to be already parsed? Ruby's URI parse (built on C) does not parse UTF8 strings, so we have a test helper that encodes the URI before it is fed to our SigV4 signer. The result was the test case was double encoded similar to what you are reporting here. However, my solution was to NOT encode the string for this test case (since it's assumed to already be encoded) and feed it into our signer - we do the sorting with the values already encoded, then join back the string. This is different than what you came up with here.

PR #107, opened 2020-10-07
Handle empty stream when computing the canonical request payload value.

wps132230:
*Issue:*
For empty payload stream in `s_build_canonical_payload()`
it will `goto on_cleanup` immediately then returns `AWS_OP_ERR`, because `aws_input_stream_read()` returns non zero for empty stream:
```c++
if (aws_input_stream_read(payload_stream, &body_buffer)) {
    goto on_cleanup;
}
```

*Description of changes:*
The fix is just do nothing with empty stream.

### DIFF
--- a/source/aws_signing.c
+++ b/source/aws_signing.c
@@ -1370,7 +1370,12 @@ static int s_build_canonical_payload(struct aws_signing_state_aws *state) {
             goto on_cleanup;
         }
 
+        int64_t payload_stream_length = 0;
         if (payload_stream != NULL) {
+            aws_input_stream_get_length(payload_stream, &payload_stream_length);
+        }
+
+        if (payload_stream_length) {
             if (aws_input_stream_seek(payload_stream, 0, AWS_SSB_BEGIN)) {
                 goto on_cleanup;
             }


source/aws_signing.c
```diff
@@ -1370,7 +1370,12 @@ static int s_build_canonical_payload(struct aws_signing_state_aws *state) {
             goto on_cleanup;
         }
 
+        int64_t payload_stream_length = 0;
         if (payload_stream != NULL) {
+            aws_input_stream_get_length(payload_stream, &payload_stream_length);
```
bretambrose, line 1375: In the general case, this is not guaranteed to succeed, although it should leave the length 0.  What happens if instead you just check the stream state?

wps132230: For empty stream, it's `operator bool()` is `true` and it's `eof()` is false - just like a non empty stream.
However, `aws_input_stream_read(payload_stream, &body_buffer)` returns `AWS_OP_ERR` for empty stream, while returns `AWS_OP_SUCCESS` for non empty stream.
The consequence is for empty payload stream, `s_build_canonical_payload` returns `AWS_OP_ERR`, `aws_signing_build_canonical_request` returns `AWS_OP_ERR`, and finally `s_perform_signing` will fail.

I encountered this issue when I assign an empty payload stream to a request in sgiv4 testing in C++SDK. And during the signing process, we will have an empty `Aws::Crt::Io::StdIOStreamInputStream`. So another possible fix is to change `StdIOStreamInputStream::ReadImpl()` in aws-crt-cpp to make It return `true` when reading from empty stream. But it doesn't help for other read implementation. 

And another possible fix for this specific usage is to replace the stream with nullptr if it's empty. But it doesn't help for a more general case, when someone else have an empty stream rather than nullptr.

Do you have any suggestions on this issue?

bretambrose: In my tests aws_input_stream_read() does not return an error when called on a either a file or memory stream.  Similar seek(Begin, 0) also succeeds for both.

wps132230: Right, that's because of different read implementations for different stream:
If reading non zero length of bytes from an empty stream, both `s_aws_input_stream_file_read` and `s_aws_input_stream_byte_cursor_read` return `AWS_OP_SUCCESS` while `Aws::Crt::Io::InputStream::s_Read` (which calls `Aws::Crt::Io::StdIOStreamInputStream::ReadImpl()`) returns `AWS_OP_ERR`.

Probably I will fix it in `Aws::Crt::Io::StdIOStreamInputStream::ReadImpl()`.

PR #109, opened 2020-10-15
Fix bug where X-Amz-Security-Token query param was added twice.

graebm:
Centralize logic so that, when we add new headers that contribute to canonical request (x-amz-date, x-amz-security-token, x-amz-content-sha-256), we also add them to signing results. Previously, the logic was scattered around, causing this bug.

Slight changes of behavior:
* We no longer add **any** X-Amz headers when doing query param signing. Previously, we were adding X-Amz-Date header (unsigned) when doing query param signing. Seems weird to add X-Amz-Date in two different places, and confirmed that botocore doesn't do this, so it's gone now.
* No longer possible to add X-Amz-Content-SHA256 header during query-param signing. We don't know of cases where a query-param signed request would even have a body. If there was a case, you'd think the service would use an X-Amz-Content-SHA256 query-param instead of a header, right?

Also, realized we were doing the `post-sts-header-before` and `post-sts-header-after` tests wrong. There's a .txt file describing what they were supposed to be testing. In the `before` test, X-Amz-Security-Token is a signed header, like normal. In the `after` test, X-Amz-Security-Token is added after signing, like with the IoT websocket upgrade request.

### DIFF
--- a/source/aws_signing.c
+++ b/source/aws_signing.c
@@ -40,6 +40,7 @@
 #define CREDENTIAL_SCOPE_STARTING_SIZE 128
 #define ACCESS_CREDENTIAL_SCOPE_STARTING_SIZE 149
 #define SCRATCH_BUF_STARTING_SIZE 256
+#define MAX_AUTHORIZATION_HEADER_COUNT 4
 #define MAX_AUTHORIZATION_QUERY_PARAM_COUNT 6
 #define DEFAULT_PATH_COMPONENT_COUNT 10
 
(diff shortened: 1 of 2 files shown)


source/aws_signing.c
```diff
@@ -1095,6 +1116,36 @@ static int s_append_canonical_header(
     return AWS_OP_SUCCESS;
 }
 
+/* Add header to stable_header_list to be canonicalized, and also to final signing result */
+static int s_add_authorization_header(
```
graebm, line 1120: Called them "authorization headers" because there was a s_add_authorization_query_param() function that does the same thing. Could rename them both s_add_x_amz_headers/query_params() or something.

bretambrose (approved):

PR #117, opened 2020-12-18
revert credential provider

TingDaoK:
*Issue #, if available:*

*Description of changes:*

- Have no unittest here, since it's should leave the provider to the language binding to implement, not sure how to unit test it.
- Naming can be changed, I really don't think `revert` is the best name, but I cannot come up with a better option...


include/aws/auth/credentials.h
```diff
@@ -406,6 +406,23 @@ struct aws_credentials_provider_chain_default_options {
     struct aws_client_bootstrap *bootstrap;
 };
 
+/**
+ * For language binding to provide credentials from that language
+ */
+struct aws_credentials_provider_revert_options {
```
bretambrose, line 412: "revert" as a descriptor doesn't seem like a good match.  "delegating" or "delegate" feel like it'd be pretty good.  "virtual" might be ok as well.   Others, like "proxy" are too overloaded with additional meanings.

include/aws/auth/credentials.h
```diff
@@ -406,6 +406,23 @@ struct aws_credentials_provider_chain_default_options {
     struct aws_client_bootstrap *bootstrap;
 };
 
+/**
+ * For language binding to provide credentials from that language
```
bretambrose, line 410: Given the generality here, there's no need to talk about language bindings.

source/crdential_provider_revert.c
```diff
@@ -0,0 +1,28 @@
...
+#include <aws/auth/credentials.h>
+
+#include <aws/auth/private/credentials_utils.h>
+#include <aws/common/environment.h>
+#include <aws/common/string.h>
+
+struct aws_credentials_provider *aws_credentials_provider_new_revert(
+    struct aws_allocator *allocator,
+    struct aws_credentials_provider_revert_options *options) {
+    struct aws_credentials_provider *provider = aws_mem_acquire(allocator, sizeof(struct aws_credentials_provider));
```
bretambrose, line 15: could calloc and drop the zero struct

source/crdential_provider_revert.c
```diff
@@ -0,0 +1,28 @@
...
+    if (provider == NULL) {
+        return NULL;
+    }
+
+    AWS_ZERO_STRUCT(*provider);
+
+    aws_credentials_provider_init_base(provider, allocator, options->provider_vtable, NULL);
```
bretambrose, line 22: pass impl instead of NULL and then drop the assignment below.

include/aws/auth/auth.h
```diff
@@ -25,6 +25,7 @@ enum aws_auth_errors {
     AWS_AUTH_SIGNING_ILLEGAL_REQUEST_HEADER,
     AWS_AUTH_SIGNING_INVALID_CONFIGURATION,
     AWS_AUTH_CREDENTIALS_PROVIDER_INVALID_ENVIRONMENT,
+    AWS_AUTH_CREDENTIALS_PROVIDER_INVALID_REVERT,
```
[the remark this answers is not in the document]
TingDaoK: It's used by the language binding, from the get_credentials function

include/aws/auth/credentials.h
TingDaoK: delegate it is!

include/aws/auth/credentials.h
TingDaoK: Yeah, changed words there

include/aws/auth/credentials.h
```diff
@@ -407,18 +407,18 @@ struct aws_credentials_provider_chain_default_options {
 };
 
 /**
- * For language binding to provide credentials from that language
+ * Configuration options for the default credentials provider chain.
```
bretambrose, line 410: delegate, and in this case it's not a chain, just a single provider.

bretambrose (approved):
Fix n ship

include/aws/auth/credentials.h
TingDaoK: whoops... copy/paste error again...

PR #133, opened 2021-06-01
Sigv4a

bretambrose:
* Sigv4a signing


include/aws/auth/signing.h
```diff
@@ -82,6 +83,53 @@ int aws_sign_request_aws(
     aws_signing_complete_fn *on_complete,
     void *userdata);
 
+/**
```
justinboswell, line 86: Normally we shove these sorts of functions into `aws/<lib>/private/` to further indicate their internal-ness. They are still installed, but only referenced in tests.

source/key_derivation.c
```diff
@@ -0,0 +1,370 @@
...
+
+    /*
+     * We only need to support comparing byte sequences of the same length here
+     */
+    const size_t lhs_len = lhs_raw_be_bigint->len;
+    if (lhs_len != rhs_raw_be_bigint->len) {
+        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+    }
+
+    volatile uint8_t gt = 0;
```
justinboswell, line 168: Should probably just add a comment here explaining why these are volatile, it's not immediately obvious unless you really think about it

include/aws/auth/signing.h
bretambrose: Actually, these are bound out in the CRTs so that SDKs can check their answers too.

source/key_derivation.c
bretambrose: line 137 explains, unsure if adding additional references would help.

justinboswell: Oh good call, I missed that.

include/aws/auth/signing.h
justinboswell: Oh. Nevermind then!

justinboswell (approved):

PR #136, opened 2021-08-03
added logging to signing validation functions

justinboswell:

### DIFF
--- a/source/aws_signing.c
+++ b/source/aws_signing.c
@@ -2372,20 +2372,29 @@ int aws_verify_sigv4a_signing(
     int result = AWS_OP_ERR;
 
     if (base_config->config_type != AWS_SIGNING_CONFIG_AWS) {
+        AWS_LOGF_ERROR(AWS_LS_AUTH_SIGNING, "Signing config is not an AWS signing config");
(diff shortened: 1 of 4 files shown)


source/aws_signing.c
```diff
@@ -2399,38 +2408,46 @@ int aws_verify_sigv4a_signing(
...
-    if (aws_credentials_get_ecc_key_pair(signing_state->config.credentials) == NULL) {
-        struct aws_credentials *ecc_credentials =
-            aws_credentials_new_ecc_from_aws_credentials(allocator, signing_state->config.credentials);
-        aws_credentials_release(signing_state->config.credentials);
-        signing_state->config.credentials = ecc_credentials;
-        if (signing_state->config.credentials == NULL) {
-            goto done;
-        }
-    }
+    struct aws_ecc_key_pair *verification_key = NULL;
```
sdavtaker, line 2411: Why not to initialize at the time of declaration? I don't see any use of verification_key until the initialization line in 2430

sdavtaker (approved):

source/aws_signing.c
justinboswell: because otherwise you get goto issues, and I don't want to touch ECC stuff until after everything else is validated

PR #144, opened 2021-10-26
Signable trailer

ilevyor:
Add support for sigv4a signing of chunked trailers


source/aws_signing.c
```diff
@@ -1715,6 +1725,45 @@ static int s_build_canonical_request_body_chunk(struct aws_signing_state_aws *st
     return AWS_OP_SUCCESS;
 }
 
+static int s_build_canonical_request_trailing_headers(struct aws_signing_state_aws *state) {
+
```
bretambrose, line 1729: Ok, so we shouldn't have followed the chunk signing pattern here.  Like request signing, this should get written to the canonical_request member, and then the hex-encoded sha256 of the "canonical request" should get written to the state via s_build_canonical_request_hash.  Obviously, drop all of the string-to-sign components out of this function.

Then in the string-to-sign construction for this signing type, we should use that value appropriately when we build the trailing header string-to-sign

source/signable_chunk.c
```diff
@@ -0,0 +1,115 @@
...
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#include <aws/auth/signable.h>
+#include <aws/common/string.h>
+
+/*
+ * This is a simple aws_signable wrapper implementation for an s3 chunk
+ */
+struct aws_signable_chunk_impl { /* <------------------ reference here -------------------- */
```
bretambrose, line 12: Remove

source/signing.c
```diff
@@ -117,12 +117,12 @@ static void s_aws_signing_on_get_credentials(struct aws_credentials *credentials
     s_perform_signing(state);
 }
 
-int aws_sign_request_aws(
-    struct aws_allocator *allocator,
-    const struct aws_signable *signable,
-    const struct aws_signing_config_base *base_config,
-    aws_signing_complete_fn *on_complete,
-    void *userdata) {
+int aws_sign_request_aws(/* <------------------------------- look here ---------------------------------- */
```
bretambrose, line 120: Remove

bretambrose (changes requested):

tests/test_chunked_signing.c
```diff
@@ -23,6 +23,7 @@
  *
  */
 
+// make sure chunk tests derives private key from here, otherwise look for chunked_tests private key source
```
bretambrose, line 26: presumably can remove

bretambrose (approved):

Issue #151, opened 2022-01-18
FreeBSD systems require -lexecinfo flag 


ring-pete:
Hello,
While trying to get the v2 aws-cli working on FreeBSD I ran into an issue where `aws` would fail due to their not being a backtrace symbol in `_awscrt.cpython-38.so`.

After doing some digging I traced this down to how cmake in this repository.  After applying the following patch I am able to successfully build and run the v2 aws-cli on my FreeBSD systems.  This will ensure FreeBSD systems use the OS provided execinfo library:

```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index b25dda8..1b35ba4 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,7 +68,7 @@ elseif (CMAKE_SYSTEM_NAME STREQUAL "Linux")
 elseif (APPLE)
     set(PLATFORM_LIBS "")
 elseif (CMAKE_SYSTEM_NAME STREQUAL "FreeBSD" OR CMAKE_SYSTEM_NAME STREQUAL "NetBSD")
-    set(PLATFORM_LIBS "")
+    set(PLATFORM_LIBS "-lexecinfo")
 endif()
 ```

I have not tested on NetBSD yet, but I believe this will be required for that platform as well.

yasminetalby:
Hello,

I am closing this issue as BSD support was fixed a few months back by @knightjoel.

For reference, as mentioned by my colleague in #152 , the fixes were in other repos. Here's some of that work:
https://github.com/awslabs/aws-c-common/pull/976
https://github.com/awslabs/aws-c-common/pull/979
https://github.com/awslabs/aws-c-common/pull/991
https://github.com/awslabs/aws-c-cal/pull/133
https://github.com/awslabs/aws-c-cal/pull/134
https://github.com/awslabs/aws-c-io/pull/553
https://github.com/awslabs/aws-c-io/pull/545

PR #152, opened 2022-01-26
Ensure BSD systems are built against OS execinfo lib

ring-pete:
*https://github.com/awslabs/aws-c-auth/issues/151*

On BSD systems we need to ensure we use the OS provided execinfo library.  Without passing this flag when the resulting cpython binary built by the awscrt will error out complaining about a missing backtrace symbol.

> Thanks for taking a look at this!
> 
> Taking a peek at `CMakeLists.txt` here: https://github.com/awslabs/aws-c-common/blob/main/CMakeLists.txt it seems like i should be including the BSD execinfo library (although I am not a cmake expert by any means):
> 
> ```
> 95    elseif(CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
> 96        list(APPEND PLATFORM_LIBS dl m thr execinfo)
> ```
> 
> I can run some tests on my end today and see if I can figure out why this isn't being picked up by aws-c-auth.

Can you also see what happens if you just drop PLATFORM_LIBS entirely from the file?


bretambrose (commented):
My initial reaction is that we shouldn't be setting PLATFORM_LIBS at all in here and that was likely a copy-and-paste error introduced years ago.

PLATFORM_LIBS should be inherited from aws-c-common.

ring-pete:
Thanks for taking a look at this!

Taking a peek at `CMakeLists.txt` here: https://github.com/awslabs/aws-c-common/blob/main/CMakeLists.txt it seems like i should be including the BSD execinfo library (although I am not a cmake expert by any means):

```
95    elseif(CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
96        list(APPEND PLATFORM_LIBS dl m thr execinfo)
```

I can run some tests on my end today and see if I can figure out why this isn't being picked up by aws-c-auth.

ring-pete:
Hello, hoping someone can revisit the PR to help get BSD support working.

graebm:
BSD support was fixed a few months back by @knightjoel (Thanks again!)

Sorry for letting this old PR sit neglected, we have lot of repos to maintain and haven't always been the best at keeping on top of external contributions.

The fixes were in other repos. Here's some of that work:
https://github.com/awslabs/aws-c-common/pull/976
https://github.com/awslabs/aws-c-common/pull/979
https://github.com/awslabs/aws-c-common/pull/991
https://github.com/awslabs/aws-c-cal/pull/133
https://github.com/awslabs/aws-c-cal/pull/134
https://github.com/awslabs/aws-c-io/pull/553
https://github.com/awslabs/aws-c-io/pull/545

ring-pete:
Thanks so much for the follow-up!  It's really appreciated and understandable how things can get lost in the shuffle.

PR #155, opened 2022-03-02
Use JSON module instead of cJSON

TwistedTwigleg:
*Description of changes:*

Modifies the code to use the JSON module in AWS C Common instead of directly using cJSON. This prevents duplicate copies/includes of cJSON, as well as helping reduce additional dependencies and code duplication.

**Note** - is dependent on the following PR: https://github.com/awslabs/aws-c-common/pull/889

___________

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.*


tests/aws_imds_client_test.c
```diff
@@ -276,6 +276,13 @@ static int s_aws_imds_tester_init(struct aws_allocator *allocator) {
     if (aws_byte_buf_init(&s_tester.resource, allocator, 256)) {
         return AWS_OP_ERR;
     }
+
+    /**
+     * Make sure the JSON module is initialized. Normally this is handled in common.c
+     * but for testing we need to initialize and clean it up manually.
+     */
+    aws_json_module_init(allocator);
```
bretambrose, line 284: auth init should eventually call common init, so this shouldn't be necessary

tests/aws_imds_client_test.c
```diff
@@ -294,6 +301,8 @@ static int s_aws_imds_tester_cleanup(void) {
 
     aws_auth_library_clean_up();
 
+    aws_json_module_cleanup();
```
bretambrose, line 304: Same

bretambrose (approved):

TwistedTwigleg:
Thanks for the reviews! Merging into `main`...

PR #156, opened 2022-03-08
cached provider: honor STS expiration time

grrtrr:
The cached credentials provider will only adjust its refresh interval when caching STS
web identity credentials (expiration_timepoint_seconds < UINT_MAX).

STS credentials are typically valid for 1h. The default refresh interval used by the default
chain provider for caching credentials is 15 minutes (`DEFAULT_CREDENTIAL_PROVIDER_REFRESH_MS`).

Currently the cached credentials provider will only adjust its refresh interval if the
STS credentials expire within its refresh interval (via if-condition).

Consequently, the cached provider will refresh STS credentials 4 times more frequently than
required, increasing the load on the STS endpoint. This causes issues for large parallel
workloads, such as throttling STS requests.

Hence honor the longer expiration interval where possible.

*Issue #, if available:*

*Description of changes:*
Let the cached credentials provider accept the expiration suggested by the cached (STS) credentials,
not the (shorter) refresh interval.


grrtrr:
_Note_: the issue of too frequent refreshes also appears with the SSO provider (which has token expiration periods of 8..12 hours). The defaults would refresh far too often here.

bretambrose:
Merged in https://github.com/awslabs/aws-c-auth/pull/166

PR #161, opened 2022-04-27
FIX: If the token is optional don't freak out

TingDaoK:
If the token is optional don't freak out


bretambrose (approved):

TwistedTwigleg (approved):
Looks good to me!
Sorry about the delay

PR #165, opened 2022-06-07
Add SSO credentials provider to the default chain

### LINKED ISSUE #164: default chain provider: add support for SSO generated credentials
The AWS C++ [SDK](https://github.com/aws/aws-sdk-cpp) supports AWS SSO credentials [since version 1.9](https://github.com/aws/aws-sdk-cpp/issues/1433#issuecomment-1079267499).

As of June 2022, the `aws-c-auth` `aws_credentials_provider_default_chain` does not support SSO credentials,
which means it can not be used on systems that rely on SSO, or called from the AWS SDK  > 1.9.0 when using the 
CommonRunTime (CRT) extension for `PutObject/GetObject`.

(Related issue: support for STS credentials #157).

grrtrr:
*Issue #, if available:* Resolves #164.

*Description of changes:*
This adds an SSO credentials provider to the default chain.


grrtrr:
> A few initial requests while holding off on a detailed review for now:

@bretambrose  - thanks for the feedback. The update to the PR so far includes/addresses:
- addressed request (2) - using `~/.aws/sso/cache` as only source directory,
- the fix uses `aws_get_home_directory` (please note https://github.com/awslabs/aws-c-common/pull/892),
- rebased with regard to recent changes,
- updated the PR so that it builds on `main` instead of `v0.6.4`.

Test cases (1) to follow.

tests/credentials_provider_sso_tests.c
```diff
@@ -0,0 +1,443 @@
...
+
+    aws_auth_library_clean_up();
+
+    return AWS_OP_SUCCESS;
+}
+AWS_TEST_CASE(
+    credentials_provider_sso_provider_access_token_expired,
+    s_credentials_provider_sso_provider_access_token_expired);
+
+static int s_credentials_provider_sso_new_destroy_from_config(struct aws_allocator *allocator, void *ctx) {
```
bretambrose, line 387: Ideally, would like to see a full success path test that uses the (admittedly clumsy) function table mocks to achieve a successful credentials fetch.

source/credentials_provider_sso.c
```diff
@@ -0,0 +1,1017 @@
...
+    if (!impl->retry_strategy) {
+        goto error;
+    }
+
+    impl->endpoint = aws_string_new_from_buf(allocator, &parameters->endpoint);
+    if (impl->endpoint == NULL) {
+        goto error;
+    }
+
+    impl->access_token = aws_string_new_from_buf(allocator, &parameters->access_token);
```
bretambrose, line 990: Ugh, I just realized after looking at the tests that we are failing provider creation on a lot of conditions that shouldn't be failures (expired token, missing token) and it's arguable that we shouldn't be failing on other situations either (profile missing, no profile file).  Instead, those should be get credentials failures.  All of this caching and calculating we're doing ahead of time should be deferred to credential query time.  If I have a long-running process, it's going to need to refresh that access token based on expiration.  And if the provider fails to create because of these conditions then even if I fix them later by refreshing or modifying the profile appropriately, it won't get used because the provider never got created.

source/credentials_provider_sso.c
```diff
@@ -0,0 +1,1017 @@
...
+        goto on_error;
+    }
+
+    if (impl->function_table->aws_http_stream_activate(stream)) {
+        goto on_error;
+    }
+
+    return;
+
+on_error:
```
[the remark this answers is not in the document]
grrtrr: I will change the assignment to `user_data->request` -
- the first call to this function has `user_data->request == NULL` (zeroed out),
- on subsequent retries `s_user_data_reset_request_and_response` sets it to `NULL`.

source/credentials_provider_sso.c
```diff
@@ -0,0 +1,1017 @@
...
+    if (aws_byte_buf_append(&access_token_path, &sha1_hex_cursor) ||
+        aws_byte_buf_append(&access_token_path, &json_cursor)) {
+        goto error;
+    }
+
+    access_token_path_str = aws_string_new_from_buf(allocator, &access_token_path);
+
+    // Use platform-specific directory separator.
+    const char local_platform_separator = aws_get_platform_directory_separator();
+    for (size_t i = 0; i < access_token_path_str->len; ++i) {
```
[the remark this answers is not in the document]
grrtrr: I was following [`aws-c-sdkutils`](https://github.com/awslabs/aws-c-sdkutils/blob/main/source/aws_profile.c#L1293-L1304).

source/credentials_provider_sso.c
```diff
@@ -0,0 +1,1017 @@
...
+
+    if (s_load_access_token_from_file(allocator, token_path, &parameters->access_token)) {
+        goto error;
+    }
+
+    if (s_construct_endpoint(allocator, aws_profile_property_get_value(sso_region),
+                             &parameters->endpoint)) {
+        goto error;
+    }
+    parameters->account_id = aws_byte_cursor_from_string(aws_profile_property_get_value(sso_account_id));
```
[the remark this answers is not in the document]
grrtrr: Good catch, I will use copies here.

source/credentials_provider_sso.c
```diff
@@ -0,0 +1,1017 @@
...
+    }
+
+    AWS_ZERO_STRUCT(*provider);
+    AWS_ZERO_STRUCT(*impl);
+
+    aws_credentials_provider_init_base(provider, allocator, &s_aws_credentials_provider_sso_vtable, impl);
+
+    aws_tls_connection_options_init_from_ctx(&tls_connection_options, options->tls_ctx);
+    /* Override the default timeout (see STS web identity provider). */
+    tls_connection_options.timeout_ms = 3 * SSO_CONNECT_TIMEOUT_DEFAULT_IN_SECONDS * 1000 /*msec*/;
```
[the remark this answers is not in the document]
grrtrr: We did run into this issue with the STS Web Identity provider, which uses a TLS negotiation timeout of 10 seconds (default).

This is one of the traces we were seeing:
```3/11/2022, 1:34:11 PM UTC	stdout	[ERROR] 2022-03-11 13:34:11.117 file-utils [140518474917760] static: Failed to open file /root/.aws/config with errno 2
3/11/2022, 1:34:11 PM UTC	stdout	[WARN] 2022-03-11 13:34:11.117 AuthProfile [140518474917760] Failed to read file at "/root/.aws/config"
3/11/2022, 1:34:11 PM UTC	stdout	[ERROR] 2022-03-11 13:34:11.117 file-utils [140518474917760] static: Failed to open file /root/.aws/credentials with errno 2
3/11/2022, 1:34:11 PM UTC	stdout	[WARN] 2022-03-11 13:34:11.117 AuthProfile [140518474917760] Failed to read file at "/root/.aws/credentials"
3/11/2022, 1:34:11 PM UTC	stdout	[ERROR] 2022-03-11 13:34:11.117 AuthCredentialsProvider [140518474917760] static: Profile credentials parser could not load or parse a credentials or config file.
3/11/2022, 1:34:21 PM UTC	stdout	[ERROR] 2022-03-11 13:34:21.164 http-connection [140518184900352] static: Client connection failed with error 1067 (AWS_IO_TLS_NEGOTIATION_TIMEOUT).
3/11/2022, 1:34:21 PM UTC	stdout	[WARN] 2022-03-11 13:34:21.164 connection-manager [140518184900352] id=0x7fcd00063b00: Failed to obtain new connection from http layer, error 1067(Channel shutdown due to tls negotiation timeout)
3/11/2022, 1:34:21 PM UTC	stdout	[WARN] 2022-03-11 13:34:21.164 connection-manager [140518184900352] id=0x7fcd00063b00: Failed to complete connection acquisition with error_code 1067(Channel shutdown due to tls negotiation timeout)
3/11/2022, 1:34:21 PM UTC	stdout	[WARN] 2022-03-11 13:34:21.164 AuthCredentialsProvider [140518184900352] id=0x7fcce7d4afe0: STS_WEB_IDENTITY provider failed to acquire a connection, error code 1067(Channel shutdown due to tls negotiation timeout)
```
The rationale for 3 x TCP connection timeout is that the TLS handshake takes 2 additional RTTs (when not using TLS 1.3).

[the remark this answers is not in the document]
grrtrr: Created #169 for reference.

source/credentials_provider_sso.c
```diff
@@ -0,0 +1,1017 @@
...
+        return NULL;
+    }
+
+    AWS_ZERO_STRUCT(*provider);
+    AWS_ZERO_STRUCT(*impl);
+
+    aws_credentials_provider_init_base(provider, allocator, &s_aws_credentials_provider_sso_vtable, impl);
+
+    aws_tls_connection_options_init_from_ctx(&tls_connection_options, options->tls_ctx);
+    /* Override the default timeout (see STS web identity provider). */
```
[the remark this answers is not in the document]
grrtrr: Will update the comment and link to #169.

tests/credentials_provider_sso_tests.c
grrtrr: Would it be possible to do that via a follow-up PR?

grrtrr: Perhaps in a future iteration?

source/credentials_provider_sso.c
grrtrr: I am not sure about this - it looks like an architectural question.
Currently the behaviour is - configuration and valid access token missing at the begin - don't use the SSO provider.

The main SDK also does not seem to have a good answer to this question yet: if the access token is missing or expired, it will just [add empty/void credentials to each request](https://github.com/aws/aws-sdk-cpp/blob/main/aws-cpp-sdk-core/source/auth/SSOCredentialsProvider.cpp#L59-L67). 
This results in error responses like the following:
```GetObject failed: HTTP response code: -1
Resolved remote host IP address: 
Request ID: 
Exception name: 
Error message: 
0 response headers:
```
It is very difficult to relate that back to a credentials problem, which is why in our code we have replaced the second return statement with `throw std::runtime_error`.

I can see a point for "access token expired" and will look into that next week. For now updating the PR with regard to above comments.

bretambrose: Well, I think a rule-of-thumb (which we may not live up to everywhere, but we don't want to make things worse) we could use is "credentials provider instantiation should not be dependent on any transient state from the process's perspective."  When part of a larger chain, it should just be a get-credentials failure that defers to the next provider in the chain.

bretambrose (commented):
Recent changes look good.  I think all that remains (which is admittedly still a bit of work) are:

1. Mocked successful query
1. Move the transient state checks that currently cause client construction failures to credentials query failures

source/credentials_provider_sso.c
grrtrr: @bretambrose - would it be possible to narrow down which failures are considered transient? I looked into the above, and I believe it would not help (we would not be able to use such an implementation).

The reason is that if the SSO provider is the only suitable credentials provider in a given environment (for instance in a k8s cluster), the precondition for the code should be that everything that the SSO provider needs to run should be there (profile set-up, up-to-date access token). 

Trying to compensate for a missing precondition in code is complex and error-prone, because if the interaction with outside processes (e.g. a long-running process to refresh an expired access token) does not run smoothly, the call to `s_aws_provider_chain_member_callback` will move on to the next provider (`ecs_or_imds_provider`), and requests will still fail with arcane error messages (the ECS or IMDS provider either failing to provide credentials, or unsuitable credentials).

Hence, would you be ok to regard the following conditions as required preconditions for the SSO provider to work?:
- `~/.aws/config` missing or malformed,
- `~/.aws/config` not containing the specified `profile`,
- Access token in `~/.aws/sso/cache/<sha1>.json` missing,
- Access token expired (since this is outside of the control of the `aws-c-auth` code and needs to be done by an external process)?

bretambrose: If you're using the default chain I don't see how the error situation is improved by skipping SSO provider creation.  In either case you'll still fully fail to source credentials.  At least if the SSO provider was present, you'd see its sourcing failure in logs.

There isn't any precedent for dynamic state causing us to drop providers entirely, failing their creation.  The closest is the conditional ecs-or-imds check.  Beyond that, the internal specification of the SSO provider implicitly precludes these failures being construction time by explicitly calling them out as potential source-time failures.

grrtrr: @bretambrose - would it meet the requirements if the SSO provider logs its failure using something like?:
```
sso: unable to configure provider (reason ...)
```

grrtrr: Since the main SDK provides [credential delegation](https://github.com/aws/aws-sdk-cpp/commit/de252525aa140cbc5822e4eeedb19f34deead66b), we are likely going to use that for SSO support.

Happy to correspond on the PR, though.

grrtrr:
Closing due to inactivity.

PR #173, opened 2022-09-06
Avoid nullptr dereference in sts creds provider client

SergeyRyabinin:
*Issue #, if available:*
Subj, avoid nullptr dereference
*Description of changes:*
Check for ptr being non-null, don't call 
```
user_data->callback(user_data->credentials, user_data->error_code, user_data->user_data);
```
(as already called in `s_clean_up_user_data`).


bretambrose (commented):
Seems fine but technically we would have crashed earlier because we're dereferencing provider_user_data from the very first line in the function.

bretambrose (approved):

SergeyRyabinin:
> Seems fine but technically we would have crashed earlier because we're dereferencing provider_user_data from the very first line in the function.

thanks for the comment!

The original finding was
```
    if (provider_user_data) {
        s_clean_up_user_data(provider_user_data);
    } else {
        provider_user_data->callback(NULL, provider_user_data->error_code, provider_user_data->user_data);
    }
```
where the ptr `provider_user_data` is first checked and in case of NULL is being immediately dereferenced to call a callback, which is obviously wrong.
Another option is to simply remove the nullptr check.

TingDaoK (approved):

waahm7 (commented):
When the provider_user_data is NULL, we never trigger the callback is wrong. I think the right solution would be to just fatal assert on provider_user_data being not NULL here, but I haven't traced through all the code paths.

PR #174, opened 2022-09-12
Cognito provider

bretambrose:
* Basic cognito provider implementation.  No identity change/update support yet.
* Updates auth testing to be a custom step where we can initialize environment variables with appropriate secrets securely
* Misc refactoring: reordered some public API declarations related to ecc-based credentials creation
* Added return values to credentials provider acquire/release per current ref-counting approach
* Extended credentials parsing helper logic to support different expiration value properties (epoch seconds vs. iso 8601 dates)
* Moved some retry strategy evaluation logic from the STS provider to a shared private space.  Updated STS provider to use the refactored APIs.
* Shared mocking function table, removed each provider's internal copy of the mock table

Note: We should eventually move to having a common simple_http_credentials_provider base object such that ECS, X509, STS, STS Web Identity, and Cognito all share the same http manager/connection/user-data and retry-related logic.  That's way out of scope for this though.


source/credentials_provider_cognito.c
```diff
@@ -0,0 +1,925 @@
...
+
+    /* return to zeroed state */
+    aws_byte_buf_clean_up(&login->login_buffer);
+    AWS_ZERO_STRUCT(*login);
+
+    return AWS_OP_ERR;
+}
+
+static void s_aws_cognito_login_clean_up(struct aws_cognito_login *login) {
+    aws_byte_buf_clean_up(&login->login_buffer);
```
graebm, line 77: trivial: AWS_ZERO_STRUCT() at the end of a clean_up() call, it's convention

source/credentials_provider_cognito.c
```diff
@@ -0,0 +1,925 @@
...
+
+    return AWS_OP_SUCCESS;
+
+on_error:
+
+    /* return to zeroed state */
+    aws_byte_buf_clean_up(&login->login_buffer);
+    AWS_ZERO_STRUCT(*login);
+
+    return AWS_OP_ERR;
```
graebm, line 73: simpler to just call s_aws_cognito_login_clean_up()?

source/credentials_provider_cognito.c
```diff
@@ -0,0 +1,925 @@
...
+    login->identity_provider_name = identity_provider_name;
+    login->identity_provider_token = identity_provider_token;
+
+    if (aws_byte_buf_append_and_update(&login->login_buffer, &login->identity_provider_name)) {
+        goto on_error;
+    }
+
+    if (aws_byte_buf_append_and_update(&login->login_buffer, &login->identity_provider_token)) {
+        goto on_error;
+    }
```
graebm, line 63: There is a helper function to do exactly this: [aws_byte_buf_init_cache_and_update_cursors()](https://github.com/awslabs/aws-c-common/blob/c815cddf645ded929f9a194b6fa6dbb1c0043427/include/aws/common/byte_buf.h#L164-L172)

source/credentials_provider_cognito.c
```diff
@@ -0,0 +1,925 @@
...
+    struct aws_input_stream *request_body_stream;
+
+    struct aws_retry_token *retry_token;
+    struct aws_credentials *credentials;
+    struct aws_byte_buf response_body;
+};
+
+static void s_user_data_reset(struct cognito_user_data *user_data) {
+    aws_byte_buf_clean_up(&user_data->request_body_buffer);
+    AWS_ZERO_STRUCT(user_data->request_body_buffer);
```
graebm, line 114: trivial: clean_up() function should have already left this struct zeroed-out

source/credentials_provider_cognito.c
```diff
@@ -0,0 +1,925 @@
...
+static struct cognito_user_data *s_user_data_new(
+    struct aws_credentials_provider *provider,
+    aws_on_get_credentials_callback_fn callback,
+    void *user_data) {
+
+    struct aws_allocator *allocator = provider->allocator;
+    struct cognito_user_data *cognito_user_data = aws_mem_calloc(allocator, 1, sizeof(struct cognito_user_data));
+    if (cognito_user_data == NULL) {
+        return NULL;
+    }
```
graebm, line 153: allocation failure is impossible now

source/credentials_provider_cognito.c
```diff
@@ -0,0 +1,925 @@
...
+
+    cognito_user_data->allocator = allocator;
+
+    if (aws_byte_buf_init(
+            &cognito_user_data->response_body, cognito_user_data->allocator, HTTP_RESPONSE_BODY_INITIAL_SIZE)) {
+        goto on_error;
```
graebm, line 159: allocation failure is impossible now

source/credentials_provider_cognito.c
```diff
@@ -0,0 +1,925 @@
...
+            "(id=%p): Cognito credentials provider failed to parse GetCredentialsForIdentity response",
+            (void *)provider_user_data->provider);
+
+        error_code = AWS_AUTH_PROVIDER_PARSER_UNEXPECTED_RESPONSE;
+    }
+
+    /* Success path is done, error-only from here on out */
+
+    /* Unsure if this should be unconditional or a function of status code. STS does this unconditionally. */
+    impl->function_table->aws_http_connection_close(provider_user_data->connection);
```
graebm, line 299: is aws_http_connection_manager_release_connection() still supposed to be called in this case?

source/credentials_provider_cognito.c
```diff
@@ -0,0 +1,925 @@
...
+        AWS_LOGF_ERROR(
+            AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+            "(id=%p): Cognito credentials provider failed to create http connection manager with error %s",
+            (void *)provider,
+            aws_error_debug_str(aws_last_error()));
+        goto on_error;
+    }
+
+    impl->endpoint = aws_string_new_from_cursor(allocator, &options->endpoint);
+    if (impl->endpoint == NULL) {
```
graebm, line 794: allocation can't fail

source/credentials_provider_cognito.c
```diff
@@ -0,0 +1,925 @@
...
+            AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+            "(id=%p): Cognito credentials provider failed to copy identity with error %s",
+            (void *)provider,
+            aws_error_debug_str(aws_last_error()));
+        goto on_error;
+    }
+
+    if (options->custom_role_arn != NULL) {
+        impl->custom_role_arn = aws_string_new_from_cursor(allocator, options->custom_role_arn);
+        if (impl->custom_role_arn == NULL) {
```
graebm, line 815: allocation can't fail

source/credentials_provider_cognito.c
```diff
@@ -0,0 +1,925 @@
...
+        AWS_LOGF_ERROR(
+            AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+            "(id=%p): Cognito credentials provider failed to copy endpoint with error %s",
+            (void *)provider,
+            aws_error_debug_str(aws_last_error()));
+        goto on_error;
+    }
+
+    impl->identity = aws_string_new_from_cursor(allocator, &options->identity);
+    if (impl->identity == NULL) {
```
graebm, line 804: allocation can't fail

source/credentials_provider_cognito.c
```diff
@@ -0,0 +1,925 @@
...
+                AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+                "(id=%p): Cognito credentials provider failed to copy custom_role_arn with error %s",
+                (void *)provider,
+                aws_error_debug_str(aws_last_error()));
+            goto on_error;
+        }
+    }
+
+    if (aws_array_list_init_dynamic(&impl->logins, allocator, options->login_count, sizeof(struct aws_cognito_login))) {
+        AWS_LOGF_ERROR(
```
graebm, line 826: allocation can't fail

source/credentials_provider_ecs.c
```diff
@@ -148,7 +137,7 @@ static void s_ecs_finalize_get_credentials_query(struct aws_credentials_provider
     struct aws_credentials *credentials = NULL;
     struct aws_parse_credentials_from_json_doc_options parse_options = {
         .access_key_id_name = "AccessKeyId",
-        .secrete_access_key_name = "SecretAccessKey",
+        .secret_access_key_name = "SecretAccessKey",
```
graebm, line 140: secretions 🤤

source/credentials_utils.c
```diff
@@ -49,6 +71,80 @@ void aws_credentials_provider_invoke_shutdown_callback(struct aws_credentials_pr
...
+
+            if (!aws_json_value_is_string(value)) {
+                AWS_LOGF_INFO(
+                    AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+                    "Expected string for type of credentials Expiration field in Json document.");
+                return false;
+            }
+
+            if (aws_json_value_get_string(value, &expiration_cursor)) {
+                AWS_LOGF_INFO(
```
graebm, line 100: trivial: the only reason aws_json_value_get_string() will will raise an error is if it's not a string, so you could omit that `if (!aws_json_value_is_string(value))` check entirely

source/credentials_utils.c
```diff
@@ -49,6 +71,80 @@ void aws_credentials_provider_invoke_shutdown_callback(struct aws_credentials_pr
...
+        case AWS_PCEF_NUMBER_UNIX_EPOCH:
+            if (!aws_json_value_is_number(value)) {
+                AWS_LOGF_INFO(
+                    AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+                    "Expected number for type of credentials Expiration field in Json document.");
+                return false;
+            }
+
+            double expiration_value = 0;
+            if (aws_json_value_get_number(value, &expiration_value)) {
```
graebm, line 133: same, just try and get the value, don't bother checking type

.builder/action/aws-c-auth-test.py
```diff
@@ -0,0 +1,44 @@
...
+               '--secret-id', secret_id]
+        # NOTE: print command args, but use "quiet" mode so that output isn't printed.
+        # we don't want secrets leaked to the build log
+        print('>', subprocess.list2cmdline(cmd))
+        result = env.shell.exec(*cmd, check=True, quiet=True)
+        secret_value = json.loads(result.output)
+        return secret_value['SecretString']
+
+    def run(self, env):
+        env.shell.setenv("AWS_TESTING_COGNITO_IDENTITY", self._get_secret(env, "aws-c-auth-testing/cognito-identity"))
```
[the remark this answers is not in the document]
bretambrose: Thought that was a Python function.  Oops.  Deleted the previous identity and switched to a new one.

.github/workflows/ci.yml
```diff
@@ -36,7 +36,7 @@ jobs:
     - name: Build ${{ env.PACKAGE_NAME }}
       run: |
         aws s3 cp s3://aws-crt-test-stuff/ci/${{ env.BUILDER_VERSION }}/linux-container-ci.sh ./linux-container-ci.sh && chmod a+x ./linux-container-ci.sh
-        ./linux-container-ci.sh ${{ env.BUILDER_VERSION }} aws-crt-${{ matrix.image }} build -p ${{ env.PACKAGE_NAME }} downstream
+        ./linux-container-ci.sh ${{ env.BUILDER_VERSION }} aws-crt-${{ matrix.image }} build -p ${{ env.PACKAGE_NAME }} --cmake-extra=-DAWS_HAS_COGNITO_ENVIRONMENT=ON downstream
```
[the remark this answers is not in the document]
bretambrose: I don't think there are any CI situations where I wouldn't want to do it so I'll try the builder args route.  I never trust the whole '+' etc... notation having encountered a number of bugs with the overlay logic.

builder.json
```diff
@@ -14,6 +14,6 @@
     "downstream": [
     ],
     "test_steps": [
-        "test"
+        "aws-c-auth-test"
```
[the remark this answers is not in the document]
bretambrose: Would have thought no originally but since env is a builder thing, it seems likely that steps that side-affect it could hold for subsequent steps so I'll give it a try.

source/credentials_provider_cognito.c
bretambrose: I still disagree sooooo much with that =/

source/credentials_provider_cognito.c
bretambrose: I just... the whole crash on allocation argument reminds me of checked exceptions.  Take this function for example.  If I remove the error checking wrt calloc/byte_buf_init then this function is now also succeed-or-crash.  So now must all the callers of this function stop checking success?  Where do you draw the line?

source/credentials_provider_cognito.c
bretambrose: Should be called in s_user_data_reset

builder.json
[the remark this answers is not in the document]
bretambrose: Tried this, but couldn't get it to work.  No idea why, as soon as I switched to two steps and renamed the steps (both here and in the python file) it just tried to run aws-c-auth-test-prep on the command line.

graebm (approved):

builder.json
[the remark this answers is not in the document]
bretambrose: Tried it again after it seemed to work on aws-crt-cpp and voila now it works.  Must have done something dumb the first time around.

PR #184, opened 2023-02-01
Fix Trailer Signing

waahm7:
*Description of changes:*
- `aws_signable_new_trailing_headers` should acquire trailer headers as it doesn't makes a deep copy of byte cursor.
- Update test to cater this use case.

### DIFF
--- a/include/aws/auth/signable.h
+++ b/include/aws/auth/signable.h
@@ -215,7 +215,7 @@ AWS_AUTH_API
 struct aws_signable *aws_signable_new_trailing_headers(
     struct aws_allocator *allocator,
     struct aws_http_headers *trailing_headers,
-    const struct aws_byte_cursor previous_signature);
+    struct aws_byte_cursor previous_signature);
 
 /**
  * Creates a signable that represents a pre-computed canonical request from an http request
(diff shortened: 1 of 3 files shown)


source/signable_trailer.c
```diff
@@ -97,12 +99,14 @@ struct aws_signable *aws_signable_new_trailing_headers(
     AWS_ZERO_STRUCT(*signable);
     AWS_ZERO_STRUCT(*impl);
 
+    aws_http_headers_acquire(trailing_headers);
```
graebm, line 102:
```suggestion
    /* Keep the headers alive. We're referencing the underlying strings. */
    aws_http_headers_acquire(trailing_headers);
```

include/aws/auth/signable.h
```diff
@@ -215,7 +215,7 @@ AWS_AUTH_API
 struct aws_signable *aws_signable_new_trailing_headers(
     struct aws_allocator *allocator,
     struct aws_http_headers *trailing_headers,
-    struct aws_byte_cursor previous_signature);
+    const struct aws_byte_cursor previous_signature);
```
graebm, line 218:
```suggestion
    struct aws_byte_cursor previous_signature);
```
the byte cursor is already passed by value/copy
there's no point in making it const

graebm (approved):
fix & ship

PR #186, opened 2023-02-10
Event Signing Support via Sigv4

waahm7:
*Description of changes:*
- Adds Sigv4 event signing support.


include/aws/auth/signing_config.h
```diff
@@ -138,7 +137,7 @@ AWS_AUTH_API extern const struct aws_byte_cursor
  * 'STREAMING-AWS4-HMAC-SHA256-EVENTS'
  * For use with `aws_signing_config_aws.signed_body_value`.
  *
- * Event signing is not yet supported
+ * Event signing is only supported for Sigv4 for now.
```
graebm, line 140: probably remove (see other comment)

source/aws_signing.c
```diff
@@ -1707,6 +1707,90 @@ static int s_build_canonical_request_hash(struct aws_signing_state_aws *state) {
     return result;
 }
 
+static int s_build_canonical_request_event(struct aws_signing_state_aws *state) {
+    int result = AWS_OP_ERR;
+
+    struct aws_byte_buf *dest = &state->string_to_sign_payload;
+
+    /* previous signature + \n */
```
graebm, line 1715: trivial: For long functions like this, it's nice to block out the major sections with bigger comment blocks. And it's nice to call back to the larger comment above...
```suggestion
    /*
     * Hex(priorSignature) + "\n"
     * 
     * Fortunately, the prior signature is already hex.
     */
```

source/aws_signing.c
```diff
@@ -1707,6 +1707,90 @@ static int s_build_canonical_request_hash(struct aws_signing_state_aws *state) {
...
+        return AWS_OP_ERR;
+    }
+
+    if (aws_byte_buf_append_byte_dynamic(dest, '\n')) {
+        return AWS_OP_ERR;
+    }
+
+    /* Append hex sha encoded date header */
+    const struct aws_signing_config_aws *config = &state->config;
+
```
graebm, line 1737: trivial: maybe just remove this variable. It's weird to be like "OK NOW WE'RE DOING THE DATE HEADER" and then make this variable and then again be like "OK NOW WE'RE DOING THE DATE HEADER AND HERE'S HOW WE'RE DOING IT" and you don't use this variable until 20 lines later
```suggestion
```

source/aws_signing.c
```diff
@@ -1707,6 +1707,90 @@ static int s_build_canonical_request_hash(struct aws_signing_state_aws *state) {
...
+    /*
+     * Encode the date header.
+     * This duplicates event stream timestamp header encoding implementation here to avoid a direct dependency.
+     * More information: https://docs.aws.amazon.com/transcribe/latest/dg/streaming-setting-up.html
+     */
```
graebm, line 1742: Nice comment, but editing so it calls back to the larger comment, and adding ascii art because that's great for understanding encodings
```suggestion
    /*
     * HexHash(nonSignatureHeaders) + "\n"
     *
     * nonSignatureHeaders is just the ":date" header.
     * We need to encode these headers in event-stream format, as described here:
     * https://docs.aws.amazon.com/transcribe/latest/dg/streaming-setting-up.html
     * 
     * | Header Name Length | Header Name | Header Value Type | Header Value Length | Header Value |
     * |       1 byte       |   N bytes   |       1 byte      |        2 bytes      |    N bytes   |
     */
```

source/aws_signing.c
```diff
@@ -1707,6 +1707,90 @@ static int s_build_canonical_request_hash(struct aws_signing_state_aws *state) {
...
+    struct aws_byte_cursor digest_cursor = aws_byte_cursor_from_buf(&digest_buffer);
+    if (aws_hex_encode_append_dynamic(&digest_cursor, dest)) {
+        goto cleanup;
+    }
+
+    if (aws_byte_buf_append_byte_dynamic(dest, '\n')) {
+        goto cleanup;
+    }
+
+    /* current hash */
```
graebm, line 1779:
```suggestion
    /*
     * HexHash(payload);
     *
     * The payload was already hashed in an earlier stage
     */
```

source/aws_signing.c
```diff
@@ -1946,6 +2030,8 @@ int aws_signing_build_canonical_request(struct aws_signing_state_aws *state) {
         case AWS_ST_CANONICAL_REQUEST_HEADERS:
         case AWS_ST_CANONICAL_REQUEST_QUERY_PARAMS:
             return s_apply_existing_canonical_request(state);
+        case AWS_ST_HTTP_REQUEST_EVENT:
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

include/aws/auth/signable.h
```diff
@@ -187,6 +187,7 @@ struct aws_signable *aws_signable_new_http_request(struct aws_allocator *allocat
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

source/aws_signing.c
waahm7: Thanks, fixed.

source/signing_config.c
```diff
@@ -52,14 +52,14 @@ int aws_validate_aws_signing_config_aws(const struct aws_signing_config_aws *con
...
     }
 
-    if (config->signature_type == AWS_ST_HTTP_REQUEST_EVENT) {
+    if (config->signature_type == AWS_ST_HTTP_REQUEST_EVENT && config->algorithm != AWS_SIGNING_ALGORITHM_V4) {
         /*
          * Not supported yet.
          *
-         * Need to determine how the (header) properties on the event signable precisely factor into the
-         * string-to-sign.  Transcribe's examples are insufficient.
+         * Need to determine if the Transcribe service supports Sigv4a and how to test it.
```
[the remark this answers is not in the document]
waahm7: Discussed offline, we will not be adding support for SigV4a for now as we don't know how to write tests for it.

include/aws/auth/signing_config.h
waahm7: Thanks, replied on the other comment.

graebm (approved):
Have Bret take a look, but LGTM!

source/aws_signing.c
```diff
@@ -1707,6 +1707,123 @@ static int s_build_canonical_request_hash(struct aws_signing_state_aws *state) {
...
+
+    if (aws_byte_buf_append_dynamic(dest, &prev_signature_cursor)) {
+        return AWS_OP_ERR;
+    }
+
+    if (aws_byte_buf_append_byte_dynamic(dest, '\n')) {
+        return AWS_OP_ERR;
+    }
+
+    /*
```
bretambrose, line 1760: Would it be worth breaking this out into a separate encode style function?

Is there any relationship between these "events" and what we have in aws-c-event-stream?

source/aws_signing.c
```diff
@@ -1707,6 +1707,123 @@ static int s_build_canonical_request_hash(struct aws_signing_state_aws *state) {
...
+    }
+
+    struct aws_byte_cursor header_name = aws_byte_cursor_from_c_str(":date");
+    AWS_FATAL_ASSERT(aws_byte_buf_write_u8(&date_buffer, (uint8_t)header_name.len));
+    if (aws_byte_buf_append_dynamic(&date_buffer, &header_name)) {
+        goto cleanup;
+    }
+
+    /* Type of timestamp header */
+    AWS_FATAL_ASSERT(aws_byte_buf_write_u8(&date_buffer, 8));
```
bretambrose, line 1786: AWS_EVENT_STREAM_HEADER_TIMESTAMP?

source/signing_config.c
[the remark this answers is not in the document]
bretambrose: Why don't we know how to write tests for it?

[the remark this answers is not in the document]
waahm7: The Transcribe documentation is insufficient and states sigv4. I tried using sigv4a in the Swift’s integration test but the server rejected it. We can add support and write tests for it in the future if the need arises. Currently, no one is asking for Sigv4a for Events.

source/aws_signing.c
waahm7: The logic of encoding date header is same as in aws-c-event-stream. I think taking a dependency on aws-c-event-stream is not worth it for a few lines encoding.

source/aws_signing.c
waahm7: Yes

bretambrose: But... every CRT includes them both as a closure.  Add the dependency.

source/aws_signing.c
bretambrose: Gonna disagree with that, given how the CRTs are constructed.

waahm7: The monolith nature of the CRT is a recurring pain point. I think we should not make the problem worse. We can reconsider this choice if this becomes more involved than encoding a single date header.

bretambrose: aws-c-event-stream is already in every CRT with the exception of JS, where it will be in soon.

source/aws_signing.c
graebm: suggestion: `, 8 /*AWS_EVENT_STREAM_HEADER_TIMESTAMP*/)`

source/aws_signing.c
waahm7: Discussed offline that we will not be taking dependency on event-stream for now.

PR #187, opened 2023-02-15
Add Option to Use Cached Profile Collection

waahm7:
*Description of changes:*
- Provides an option to use cached profile & config profiles to avoid duplicate loading of files.


source/credentials_provider_profile.c
```diff
@@ -45,52 +46,60 @@ static int s_profile_file_credentials_provider_get_credentials_async(
...
+                "(id=%p) Profile credentials provider successfully built config profile collection from file at (%s)",
+                (void *)provider,
+                aws_string_c_str(impl->config_file_path));
+        } else {
+            AWS_LOGF_DEBUG(
+                AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+                "(id=%p) Profile credentials provider failed to build config profile collection from file at (%s)",
+                (void *)provider,
+                aws_string_c_str(impl->config_file_path));
+        }
```
sbSteveK, line 70: Should we log that we are using merged_profiles (profile_collection_cached) and its values here?

include/aws/auth/credentials.h
```diff
@@ -111,6 +111,12 @@ struct aws_credentials_provider_profile_options {
      */
     struct aws_byte_cursor credentials_file_name_override;
 
+    /**
+     * Use a cached merged profile collection.
```
TingDaoK, line 115: `a cached merged profile collection`

I was really confused by this, until I read through the code. So, maybe worth to clarify that the merged means it has both config-file-based profile collection and a credentials-file-based profile collection?

include/aws/auth/credentials.h
```diff
@@ -330,6 +336,11 @@ struct aws_credentials_provider_sts_web_identity_options {
      */
     struct aws_client_bootstrap *bootstrap;
 
+    /**
+     * Use a cached config profile collection.
```
TingDaoK, line 340: why this is `cached config profile collection.` instead of merged?

source/credentials_provider_profile.c
```diff
@@ -45,52 +46,60 @@ static int s_profile_file_credentials_provider_get_credentials_async(
...
+        aws_profile_collection_release(credentials_profiles);
     }
 
-    /*
-     * Merge the (up to) two sources into a single unified profile
-     */
-    struct aws_profile_collection *merged_profiles =
-        aws_profile_collection_new_from_merge(provider->allocator, config_profiles, credentials_profiles);
     if (merged_profiles != NULL) {
         const struct aws_profile *profile = aws_profile_collection_get_profile(merged_profiles, impl->profile_name);
```
TingDaoK, line 104: Maybe I didn't fully understand the profile collection, but, how can we tell the profile collection provided is a "merged" profile collection or not to verify that people has the right config passed in?

waahm7: It is up to the customers to pass a correct collection which contains everything we need. If it doesn't, we will throw a source failure error.

include/aws/auth/credentials.h
waahm7: Thanks, fixed.

source/credentials_provider_profile.c
TingDaoK: > If it doesn't, we will throw a source failure error.

When we throw the failure? How will we throw it?

source/credentials_provider_profile.c
```diff
@@ -359,40 +372,44 @@ struct aws_credentials_provider *aws_credentials_provider_new_profile(
...
 
-    if (!(config_profiles || credentials_profiles)) {
-        AWS_LOGF_ERROR(
-            AWS_LS_AUTH_CREDENTIALS_PROVIDER,
-            "static: Profile credentials parser could not load or parse"
-            " a credentials or config file.");
-        goto on_finished;
-    }
+    merged_profiles = aws_profile_collection_acquire(options->profile_collection_cached);
+    if (!merged_profiles) {
```
graebm, line 383: trivial: suggestion is more lines of code, but IMO reads more clearly to a newcomer
```suggestion
    if (options->profile_collection_cached) {
        /* Use cached profile collection */
        merged_profiles = aws_profile_collection_acquire(options->profile_collection_cached);
    } else {
        /* Load profile collection from files */

```

source/credentials_provider_profile.c
```diff
@@ -359,40 +372,44 @@ struct aws_credentials_provider *aws_credentials_provider_new_profile(
...
+
+        if (!(config_profiles || credentials_profiles)) {
+            AWS_LOGF_ERROR(
+                AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+                "static: Profile credentials parser could not load or parse"
+                " a credentials or config file.");
+            goto on_finished;
+        }
 
+        merged_profiles = aws_profile_collection_new_from_merge(allocator, config_profiles, credentials_profiles);
```
graebm, line 411: possible bug in both old and new code ....

if only 1/2 of the files is successfully loaded, we still call `aws_profile_collection_new_from_merge()`

not sure what the idea behavior is:
- is it OK to only load 1/2 files?
- if it's OK, should we silently ignore the error like we do now, even when a filepath override is passed in?
- is it OK to call merge when 1/2 profiles is NULL?

source/credentials_provider_sts_web_identity.c
```diff
@@ -982,13 +984,15 @@ static struct sts_web_identity_parameters *s_parameters_new(struct aws_allocator
...
+    struct aws_profile_collection *config_profile = aws_profile_collection_acquire(config_profile_collection_cached);
     struct aws_string *profile_name = NULL;
     const struct aws_profile *profile = NULL;
     bool get_all_parameters =
         (region && region->len && role_arn && role_arn->len && token_file_path && token_file_path->len);
     if (!get_all_parameters) {
-        config_profile = s_load_profile(allocator);
+        if (!config_profile) {
+            config_profile = s_load_profile(allocator);
+        }
```
graebm, line 995: trivial/debatable: similar to earlier feedback, I feel like  newcomer to this code would have an easier time reading something like 
```
if (config_profile_collection_cached) {
    /* Use cached profile collection */
    config_profile = aws_profile_collection_acquire(config_profile_collection_cached);
} else {
    /* Load profile collection from files */
    config_profile = s_load_profile(allocator);
}
```

graebm: do we need to check the results of `s_load_profile()`?

oh ... a bunch of lines later we catch the error, but it prints out the same message as if profile-load succeeded and but get-profile-by-name failed

maybe better to `goto on_finish;` earlier?

source/credentials_provider_profile.c
waahm7: I think it is okay to load only 1/2 of the files. We will attempt to load properties from the file and generate an error if any necessary property is missing. In `aws_profile_collection_new_from_merge`, we do handle 1/2 of the files.

```
    if (config_profiles != NULL) {
        if (s_profile_collection_merge(merged, config_profiles)) {
            AWS_LOGF_ERROR(AWS_LS_SDKUTILS_PROFILE, "Failed to merge config profile set");
            goto cleanup;
        }
    }

    if (credentials_profiles != NULL) {
        if (s_profile_collection_merge(merged, credentials_profiles)) {
            AWS_LOGF_ERROR(AWS_LS_SDKUTILS_PROFILE, "Failed to merge credentials profile set");
            goto cleanup;
        }
    }
```

source/credentials_provider_profile.c
waahm7: We will throw an error if anything is missing from merged collection. Like [profile](https://github.com/awslabs/aws-c-auth/blob/8308060ab0d908a6242afdf369f6ea2d30a08e22/source/credentials_provider_profile.c#L95) is missing or credentials are missing https://github.com/awslabs/aws-c-auth/blob/8308060ab0d908a6242afdf369f6ea2d30a08e22/source/aws_profile.c#L34.

source/credentials_provider_sts_web_identity.c
waahm7: Thanks, updated.

graebm (approved):

PR #189, opened 2023-03-13
Implement SSOCredentialsProvider

waahm7:
Issue #164 

*Description of changes:*
- Implements SSOCredentialsProvider without the token refresh functionality.


source/token_provider_sso_profile.c
```diff
@@ -0,0 +1,184 @@
...
+/*
+ * sso-token profile provider implementation
+ */
+struct aws_token_provider_profile_impl {
+    struct aws_string *token_file_path;
+
+    aws_io_clock_fn *system_clock_fn;
+};
+
+static int s_token_provider_profile_get_token_async(
```
TingDaoK, line 28: why naming it `*_async`, there is no async call in here?

source/token_provider_sso_profile.c
```diff
@@ -0,0 +1,184 @@
...
+    if (!credentials) {
+        AWS_LOGF_ERROR(AWS_LS_AUTH_CREDENTIALS_PROVIDER, "(id=%p) Unable to construct credentials.", (void *)provider);
+        goto done;
+    }
+
+    callback(credentials, AWS_OP_SUCCESS, user_data);
+    result = AWS_OP_SUCCESS;
+
+done:
+    aws_sso_token_destroy(sso_token);
```
TingDaoK, line 69: make sure to invoke callback on failure as well

The callback should be invoked for both failure and success.

source/token_provider_sso_profile.c
```diff
@@ -0,0 +1,184 @@
...
+        allocator,
+        2,
+        &provider,
+        sizeof(struct aws_credentials_provider),
+        &impl,
+        sizeof(struct aws_token_provider_profile_impl));
+    AWS_ZERO_STRUCT(*provider);
+    AWS_ZERO_STRUCT(*impl);
+    aws_credentials_provider_init_base(provider, allocator, &s_aws_token_provider_profile_vtable, impl);
+    impl->token_file_path = aws_string_new_from_string(allocator, token_path);
```
TingDaoK, line 175: Trivial: we can avoid this copy by allocating *impl early, just need to release when the token_path failed to construct.

source/token_provider_sso_session.c
```diff
@@ -0,0 +1,229 @@
...
+
+/*
+ * sso-session token provider implementation
+ */
+struct aws_token_provider_sso_session_impl {
+    struct aws_string *token_file_path;
+    aws_io_clock_fn *system_clock_fn;
+};
+
+static int s_token_provider_sso_session_get_token_async(
```
TingDaoK, line 27: ditto: all those similar comments from the `s_token_provider_profile_get_token_async`

source/token_provider_sso_session.c
```diff
@@ -0,0 +1,229 @@
...
+
+struct aws_credentials_provider *aws_token_provider_new_sso_session(
+    struct aws_allocator *allocator,
+    const struct aws_token_provider_sso_session_options *options) {
+
+    /* Currently, they are not used but they will be required when we implement the refresh token functionality. */
+    AWS_ASSERT(options->bootstrap);
+    AWS_ASSERT(options->tls_ctx);
+
+    struct aws_string *token_path = s_verify_config_and_construct_token_path(
```
TingDaoK, line 201: ditto: avoid a separate token_path by using the one from impl.

source/token_provider_sso_profile.c
waahm7: The result will be AWS_OP_ERR, so we should not invoke the callback.

source/token_provider_sso_session.c
```diff
@@ -0,0 +1,229 @@
...
+        AWS_LOGF_ERROR(
+            AWS_LS_AUTH_CREDENTIALS_PROVIDER, "sso-session: token parser failed to find sso_start_url in sso-session");
+        aws_raise_error(AWS_AUTH_SSO_TOKEN_PROVIDER_SOURCE_FAILURE);
+        goto cleanup;
+    }
+
+    const struct aws_string *sso_region = aws_profile_property_get_value(sso_region_property);
+    const struct aws_string *sso_start_url = aws_profile_property_get_value(sso_start_url_property);
+
+    /* Verify sso_region & start_url are the same in profile section if they exist */
```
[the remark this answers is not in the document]
waahm7: Yes, they should have the same value if they exist.

source/sso_token_utils.c
```diff
@@ -0,0 +1,165 @@
...
+#include <aws/common/file.h>
+#include <aws/common/json.h>
+
+#if defined(_MSC_VER)
+#    pragma warning(disable : 4232)
+#endif /* _MSC_VER */
+
+AWS_STATIC_STRING_FROM_LITERAL(s_sso_cache_directory, "/.aws/sso/cache/");
+
+struct aws_string *aws_construct_token_path(struct aws_allocator *allocator, const struct aws_string *input) {
```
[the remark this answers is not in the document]
waahm7: I have renamed it to `aws_construct_sso_token_path`. In general, `token` is a generic term which is just a string. It can be anything like accessToken, sessionToken, bearer token etc. `sso_token` is a json used by the sso service, which contains the following 
```
{
    "accessToken": " ",
    "expiresAt": "",
    "refreshToken": "string",
    "clientId": " ",
    "clientSecret": " ",
    "registrationExpiresAt": " ",
    "region": " ",
    "startUrl": " "
}
```
Currently, we are only using accessToken out of these fields thus they look similar, but when we implement the refresh functionality, we will need other things like refresh token.

source/token_provider_sso_profile.c
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    /* check token expiration. */
+    uint64_t now_ns = UINT64_MAX;
+    if (impl->system_clock_fn(&now_ns) != AWS_OP_SUCCESS) {
+        goto done;
+    }
+    uint64_t now_seconds = aws_timestamp_convert(now_ns, AWS_TIMESTAMP_NANOS, AWS_TIMESTAMP_SECS, NULL);
+
+    if (aws_date_time_as_epoch_secs(&sso_token->expiration) - now_seconds < 0) {
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

source/credentials_utils.c
```diff
@@ -292,3 +294,43 @@ enum aws_retry_error_type aws_credentials_provider_compute_retry_error_type(int
...
+            aws_error_str(aws_last_error()));
+        goto on_error;
+    }
+
+    aws_string_destroy(config_file_path);
+    return config_profiles;
+
+on_error:
+    aws_string_destroy(config_file_path);
+    aws_profile_collection_destroy(config_profiles);
```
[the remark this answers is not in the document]
waahm7: thanks, fixed.

source/sso_token_utils.c
```diff
@@ -0,0 +1,165 @@
...
+            AWS_LS_AUTH_CREDENTIALS_PROVIDER, "sso token: failed to load token file %s", aws_string_c_str(file_path));
+        goto cleanup;
+    }
+
+    struct aws_byte_cursor document_cursor = aws_byte_cursor_from_buf(&file_contents_buf);
+    document_root = aws_json_value_new_from_string(allocator, document_cursor);
+    if (document_root == NULL) {
+        AWS_LOGF_ERROR(
+            AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+            "sso token: failed to parse access token file %s",
```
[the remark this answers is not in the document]
waahm7: Thanks, updated the message.

source/token_provider_sso_profile.c
waahm7: This is debatable because SSOCredentialsProvider will be added to default chain. Here are the three options that I could think of.

1. Allocate impl early and release it if token_path creation failed. As this will be part of default chain, It will be frequently visited path that we try to create SSOCredentialsProvider but the config is not right. This will result in unnecessary allocations.
2. We could simply pass the token by reference to the impl instead of making a deep copy and destroying it. This is the most optimal but suffers from readability as the allocation and deallocaiton are not on the same level.
3. Current approach which does an extra allocation for token path.

source/token_provider_sso_profile.c
waahm7: Thanks, fixed.

source/token_provider_sso_session.c
```diff
@@ -0,0 +1,229 @@
...
+    if (!sso_session_property) {
+        AWS_LOGF_ERROR(
+            AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+            "sso-session: token provider could not find an sso-session at profile %s",
+            aws_string_c_str(profile_name));
+        aws_raise_error(AWS_AUTH_SSO_TOKEN_PROVIDER_SOURCE_FAILURE);
+        goto cleanup;
+    }
+    const struct aws_string *sso_session_name = aws_profile_property_get_value(sso_session_property);
+    const struct aws_profile *session_profile =
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

source/sso_token_utils.c
```diff
@@ -0,0 +1,165 @@
...
+        goto cleanup;
+    }
+
+    // Use platform-specific directory separator.
+    const char local_platform_separator = aws_get_platform_directory_separator();
+    for (size_t i = 0; i < token_path_buf.len; ++i) {
+        if (aws_is_any_directory_separator((char)token_path_buf.buffer[i])) {
+            ((char *)token_path_buf.buffer)[i] = local_platform_separator;
+        }
+    }
```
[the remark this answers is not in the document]
waahm7: I created the [PR](https://github.com/awslabs/aws-c-common/pull/1010) to normalize paths, but it crashes if someone passes non-writable memory to the function. We can always allocate memory for normalized path but that is not that efficient. I realized that we can just construct a cross-platform path instead of constructing then normalizing.

include/aws/auth/credentials.h
```diff
@@ -640,6 +640,20 @@ struct aws_credentials *aws_credentials_new_ecc_from_aws_credentials(
     struct aws_allocator *allocator,
     const struct aws_credentials *credentials);
 
+/**
+ * Creates a set of AWS credentials based on a token with expiration.
```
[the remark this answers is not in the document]
TingDaoK: I thought we were making the whole "token credential" thing private

include/aws/auth/private/sso_token_providers.h
```diff
@@ -0,0 +1,97 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#include <aws/auth/auth.h>
+#include <aws/auth/credentials.h>
+
+/**
+ * Configuration options for a provider that sources sso token information from the aws profile (by default
+ * ~/.aws/config) and token from ~/.aws/sso/cache/<sha1 of start url>.json.
```
TingDaoK, line 14: Worth to document this is the legacy way?

source/credentials_utils.c
```diff
@@ -3,8 +3,10 @@
  * SPDX-License-Identifier: Apache-2.0.
  */
 
+#include <aws/auth/private/aws_profile.h>
```
TingDaoK, line 6: `#include <aws/auth/private/aws_profile.h>` vs `#include <aws/sdkutils/aws_profile.h>`?

source/credentials_utils.c
```diff
@@ -292,3 +294,37 @@ enum aws_retry_error_type aws_credentials_provider_compute_retry_error_type(int
 
     return error_type;
 }
+
+struct aws_profile_collection *aws_load_profile_collection_from_config_file(
```
TingDaoK, line 298: should this function be moved to `aws/sdkutils/aws_profile.h`?

include/aws/auth/private/sso_token_providers.h
waahm7: Documented as public documentation also mentions that this is the legacy way.

include/aws/auth/credentials.h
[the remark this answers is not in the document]
waahm7: Thanks, discussed offline. I have moved it to a private header.

TingDaoK (approved):

PR #193, opened 2023-04-10
Implement SSOCredentialsProvider part 3

waahm7:
*Description of changes:*
- Allows cached config to use with SSOCredentialsProvider


source/credentials_provider_sso.c
```diff
@@ -677,14 +684,15 @@ static struct sso_parameters *s_parameters_new(
...
-                config_profile,
+                config_profile_collection,
                 AWS_PROFILE_SECTION_TYPE_SSO_SESSION,
                 aws_profile_property_get_value(sso_session_property)),
             s_sso_region);
     } else {
         /* construct profile token provider */
         struct aws_token_provider_sso_profile_options token_provider_options = {
             .config_file_name_override = options->config_file_name_override,
+            .config_file_cached = options->config_file_cached,
```
TingDaoK, line 695: why not use `config_profile_collection`? Even if the options doesn't set `config_file_cached`, it's still loaded here and can be reused from token provider, right?

TingDaoK (approved):

PR #202, opened 2023-06-20
delete unused test set

TingDaoK:
The test sets has been replaced by `tests/aws-signing-test-suite/v4` and not being used, I don't see a reason to keep them around. Unless we have down streams still uses them (?)


graebm:
did it here instead: https://github.com/awslabs/aws-c-auth/pull/223

PR #207, opened 2023-07-13
Adapt to XML API changes

graebm:
- Adapt to API changes from: https://github.com/awslabs/aws-c-common/pull/1043
- Break up node traversal functions, to ensure we're processing the correct XML elements.
    - Previously, the same callback would be used for all XML elements. This could cause error if an element with the same name occurred at different parts of the document tree.
- Improved error checking
    - Previously, many calls to `aws_xml_node_as_body()` weren't being checked for error.

### DIFF
--- a/source/credentials_provider_sts_web_identity.c
+++ b/source/credentials_provider_sts_web_identity.c
@@ -37,10 +37,10 @@
 #define STS_WEB_IDENTITY_MAX_ATTEMPTS 3
 
 static void s_on_connection_manager_shutdown(void *user_data);
-static int s_stswebid_error_xml_on_Error_child(struct aws_xml_node *, bool *, void *);
-static int s_stswebid_200_xml_on_AssumeRoleWithWebIdentityResponse_child(struct aws_xml_node *, bool *, void *);
-static int s_stswebid_200_xml_on_AssumeRoleWithWebIdentityResult_child(struct aws_xml_node *, bool *, void *);
-static int s_stswebid_200_xml_on_Credentials_child(struct aws_xml_node *, bool *, void *);
+static int s_stswebid_error_xml_on_Error_child(struct aws_xml_node *, void *);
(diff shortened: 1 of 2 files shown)


source/credentials_provider_sts_web_identity.c
```diff
@@ -187,128 +191,138 @@ Error Response looks like:
...
     AWS_ZERO_STRUCT(credential_data);
+
     if (aws_byte_cursor_eq_c_str_ignore_case(&node_name, "AccessKeyId")) {
-        aws_xml_node_as_body(parser, node, &credential_data);
-        query_user_data->access_key_id =
-            aws_string_new_from_array(query_user_data->allocator, credential_data.ptr, credential_data.len);
+        if (aws_xml_node_as_body(node, &credential_data)) {
+            return AWS_OP_ERR;
+        }
+        query_user_data->access_key_id = aws_string_new_from_cursor(query_user_data->allocator, &credential_data);
```
TingDaoK, line 286: trivial: should we check the `access_key_id` is not empty? If it's empty, it will go all the way to when we create credentials to find out the error and report error while creating the credentials. So, maybe better to check it here and fail the parsing.

TingDaoK (approved):

source/credentials_provider_sts_web_identity.c
graebm: I made it so we report `AWS_AUTH_CREDENTIALS_PROVIDER_STS_WEB_IDENTITY_SOURCE_FAILURE` if anything goes wrong parsing the response, and/or creating the credentials. That seems more useful than the AWS_ERROR_INVALID_ARGUMENT we would get with the old path, if the strings were ever empty.

graebm: I decided to check at the end, that the credentials were successfully created, vs checking each element's string is non-empty. That way, if it ever becomes valid for one of these strings to be empty, we only have to fixed `aws_credentials_new()`, vs needing to fix everywhere that reads an XML element that eventually feeds into `aws_credentials_new()`

PR #211, opened 2023-08-07
Fix ProcessCredentialsProvider Callback to trigger once

waahm7:
*Description of changes:*
- Fix the function so that it does not trigger the callback and returns an error. Always return success, as we will always trigger the callback in every case. This is similar to the existing static and environment credentials provider.

### DIFF
--- a/source/credentials_provider_process.c
+++ b/source/credentials_provider_process.c
@@ -72,6 +72,8 @@ static int s_get_credentials_from_process(
         (void *)provider);
 
 on_finish:
+
+    ;
     int error_code = AWS_ERROR_SUCCESS;
     if (credentials == NULL) {
         error_code = aws_last_error();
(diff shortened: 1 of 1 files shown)


sbSteveK (approved):

TingDaoK (commented):
Why we always trigger the callback? Why not just return error and not trigger the callback?

I feel "the function failed, but I'll tell you the function succeed, go check the callback instead" is a bit awkward

waahm7:
> Why we always trigger the callback? Why not just return error and not trigger the callback?
> 
> I feel "the function failed, but I'll tell you the function succeed, go check the callback instead" is a bit awkward

It doesn't really matter which one. The main thing is that we should either trigger the callback or return an error. I choose the callback because it makes it a bit more consistent with the other credentials provider. The function succeeding just means that we were able to successfully make the callback (with credentials or an error), and error means that something went wrong and we can't even trigger the callback.

PR #215, opened 2023-09-19
StsWebIdentity allow parameters

waahm7:
*Issue #, if available:*
https://github.com/awslabs/aws-crt-swift/issues/175

*Description of changes:*
Adds an option to pass required values via parameters.


source/credentials_provider_sts_web_identity.c
```diff
@@ -934,6 +934,20 @@ static int s_generate_uuid_to_buf(struct aws_allocator *allocator, struct aws_by
     return AWS_OP_SUCCESS;
 }
 
+static void s_check_or_get_with_env(
```
TingDaoK, line 937: trivial: why not just return `aws_string *` instead of passing the `aws_string **out`?

TingDaoK (approved):

source/credentials_provider_sts_web_identity.c
waahm7: Thanks, updated.

PR #217, opened 2023-11-02
fix the imds client endless retry

TingDaoK:
*Issue #, if available:*

- For IMDS client, we don't check for token expire, instead we rely on the server response of 401 to inform the token is invalid and try to fetch the token again after the 401 received
- Also, if we failed to fetch the token for some reason, we have a fallback to insecurity path to fetch imds without token.
- So, if we failed to fetch the token, we will try to fetch wtihout token, and if the token is required, server will also response 401 for request without token, and then we end up with endless retry

*Description of changes:*
- 401 response of fetching resource should fail the request
- Instead of rely on 401 to inform us the token expired, we memorize the cached token expiration timestamp, and if it's closer to the time, we stop using the cached token


source/aws_imds_client.c
```diff
@@ -28,7 +28,7 @@
 #define IMDS_RESPONSE_TOKEN_SIZE_INITIAL 64
 #define IMDS_RESPONSE_SIZE_LIMIT 65535
 #define IMDS_CONNECT_TIMEOUT_DEFAULT_IN_SECONDS 2
-#define IMDS_DEFAULT_RETRIES 1
```
waahm7, line 31: What is the default chain timeout now when it is not running on EC2 without any credentials?

source/aws_imds_client.c
```diff
@@ -506,16 +499,30 @@ static void s_client_on_token_response(struct imds_user_data *user_data) {
      * token in following requests.
      */
     if (user_data->status_code != AWS_HTTP_STATUS_CODE_200_OK || user_data->current_result.len == 0) {
-        s_update_token_safely(user_data->client, NULL, false);
+        AWS_LOGF_DEBUG(
+            AWS_LS_IMDS_CLIENT,
+            "(id=%p) IMDS client failed to fetch token for requester %p, fall back to insecure path for the same "
```
waahm7, line 504: I am not sure if the `insure` terminology should be in our logs. Maybe we should use `v1` instead? In the future, if there is an imdsv3, this could be confusing.

source/aws_imds_client.c
```diff
@@ -506,16 +499,30 @@ static void s_client_on_token_response(struct imds_user_data *user_data) {
...
+            "requester. Received response status code: %d",
+            (void *)user_data->client,
+            (void *)user_data,
+            user_data->status_code);
+        s_update_token_safely(user_data->client, NULL, false, 0 /*expire_timestamp*/);
     } else {
         struct aws_byte_cursor cursor = aws_byte_cursor_from_buf(&(user_data->current_result));
         aws_byte_cursor_trim_pred(&cursor, aws_char_is_space);
         aws_byte_buf_reset(&user_data->imds_token, true /*zero contents*/);
         if (aws_byte_buf_append_and_update(&user_data->imds_token, &cursor)) {
```
waahm7, line 514: I was looking through this code and was wondering if the `aws_byte_buf_append_and_update` can fail? We can probably just get rid of this if fails branch.

source/aws_imds_client.c
```diff
@@ -607,15 +615,50 @@ static void s_query_complete(struct imds_user_data *user_data) {
...
     }
 
-    /* In this case we fallback to the secure imds flow. */
     if (user_data->status_code == AWS_HTTP_STATUS_CODE_401_UNAUTHORIZED) {
-        s_invalidate_cached_token_safely(user_data);
-        s_reset_scratch_user_data(user_data);
-        aws_retry_token_release(user_data->retry_token);
-        if (s_get_resource_async_with_imds_token(user_data)) {
-            s_user_data_release(user_data);
+        /* In this case we fail the request and switch to secure path for following requests. */
```
waahm7, line 619: Need to update `credentials.h` doc accordingly.

source/aws_imds_client.c
```diff
@@ -607,15 +615,50 @@ static void s_query_complete(struct imds_user_data *user_data) {
...
+            }
+        } else {
+            /* Not expected, the token was told as invalid from server */
+            AWS_LOGF_ERROR(
+                AWS_LS_IMDS_CLIENT,
+                "(id=%p) IMDS client failed to fetch resource. Server response 401 UNAUTHORIZED indicates invalid "
+                "token. requester %p.",
+                (void *)user_data->client,
+                (void *)user_data);
+            user_data->error_code = AWS_AUTH_IMDS_CLIENT_TOKEN_INVALID;
```
waahm7, line 646: If it is not expected, maybe an assert?

source/aws_imds_client.c
```diff
@@ -607,15 +615,50 @@ static void s_query_complete(struct imds_user_data *user_data) {
...
+            client->token_state = AWS_IMDS_TS_INVALID;
+            AWS_LOGF_DEBUG(
+                AWS_LS_IMDS_CLIENT,
+                "(id=%p) IMDS client's cached token is set to be invalid by requester %p.",
+                (void *)client,
+                (void *)user_data);
+        }
+        /* Flip the flag to let following requests use token as it's required. */
+        client->token_required = true;
+        aws_mutex_unlock(&client->token_lock);
```
waahm7, line 661: Discussed offline that we have the following options 

- Maintain the old behavior of attempting at least one v1 request.
-  Deprecate IMDsv1 if the Mountpoint does not require it but they are GA now so this might not be worth a breaking change.
- Perhaps introduce a setter for token required or some sort of environment variable, allowing customers to configure it themselves rather than us changing it in the background without their knowledge and hoping they will retry the request.

source/aws_imds_client.c
TingDaoK: I don't want to fix this, there are too many stuff in the codebase that we should fix...

I gave up on this rabbit hole and left todos for reference if we decide to revamp the client.

source/aws_imds_client.c
TingDaoK: I feel like it's still possible that server starts to don't like our cache token for some other reasons, it's probably not worth to crash the program

source/aws_imds_client.c
TingDaoK: reverted the change, was trying to follow other SDKs

source/aws_imds_client.c
```diff
@@ -221,6 +217,8 @@ struct imds_user_data {
      * will be adapted according to response.
      */
     bool imds_token_required;
+    /* Indicate the request is a fallback from a failure call. */
+    bool is_fallback;
```
waahm7, line 221:
```suggestion
    bool is_fallback_request;
```

source/aws_imds_client.c
```diff
@@ -607,17 +626,61 @@ static void s_query_complete(struct imds_user_data *user_data) {
...
-        if (s_get_resource_async_with_imds_token(user_data)) {
-            s_user_data_release(user_data);
+        struct aws_imds_client *client = user_data->client;
+        aws_mutex_lock(&client->token_lock);
+        if (aws_byte_buf_eq(&user_data->imds_token, &client->cached_token)) {
+            /* If the token used matches the cached token, that means the cached token is invalid. */
+            client->token_state = AWS_IMDS_TS_INVALID;
+            AWS_LOGF_DEBUG(
+                AWS_LS_IMDS_CLIENT,
+                "(id=%p) IMDS client's cached token is set to be invalid by requester %p.",
```
waahm7, line 637:
```suggestion
                "(id=%p) IMDS client's cached token is invalidated by requester %p.",
```

source/aws_imds_client.c
```diff
@@ -607,17 +626,61 @@ static void s_query_complete(struct imds_user_data *user_data) {
...
+                (void *)client,
+                (void *)user_data);
+        }
+        /* Flip the flag to let following requests use token as it's required. */
+        client->token_required = true;
```
waahm7, line 642:
```suggestion
        /* let following requests use token as it's required. */
        client->token_required = true;
```

source/aws_imds_client.c
```diff
@@ -607,17 +626,61 @@ static void s_query_complete(struct imds_user_data *user_data) {
...
+        aws_mutex_unlock(&client->token_lock);
+
+        if (!user_data->imds_token_required) {
```
waahm7, line 645: code clarity: combine the error cases into one with a comment
```suggestion
        if (!user_data->imds_token_required && !user_data->is_fallback) {
        	// falback to v1
        } else {
        	// error out 
        
        }
```

waahm7 (approved):

PR #224, opened 2024-01-10
assumeRole chain support

waahm7:
*Description of changes:*
Adds support for STS assumeRole chaining.
**Rules:**
- When chaining assume role profiles, SDKs must terminate the chain as soon as they hit a profile with static credentials even if the profile also contains a `source_profile`.
- A profile is allowed to use itself as a source profile for static credentials.

**Examples:**
1. `ProfileA` contains static creds 
    - Return credentials from the profile
2. `ProfileA` contains static creds as well as source profile set to `ProfileA`
    - Use static creds in `ProfileA` to call the sts assumeRole and returns the credentials from the sts assumeRole call.
3. `ProfileA` contains `source_profile` set to `ProfileB`. `ProfileB` contains static creds.
    - Use static creds in `ProfileB` to call the sts assumeRole and return the credentials from the sts assumeRole call.
4. `ProfileA` contains `source_profile` set to `ProfileB`. `ProfileB` contains `source_profile` set to `ProfileC`. `ProfileC` contains static creds.
    - Use static creds in `ProfileC` to call the sts assumeRole with `role_arn` from `profileB`. 
    - Use the response creds to call sts assumeRole again with `role_arn` from `profileA`. 
    - Return creds from the second STS call.
5. `ProfileA` contains `source_profile` set to `ProfileB`. `ProfileB` contains `source_profile` set to `ProfileC` and static creds. `ProfileC` contains static creds.
    - Use static creds in `ProfileB` to call the sts assumeRole with `role_arn` from `profileA`.
    - Return creds from the STS call.
6. `ProfileA` contains `source_profile` set to `ProfileB`. `ProfileB` contains `source_profile` set to `ProfileA` with/without static creds.
    - Error out with profile contains a cycle
7. `ProfileA` contains `source_profile` set to `ProfileB`. `ProfileB` contains `source_profile` set to `ProfileC` and also contains partial static credentials. `Profile C` contains credentials.
    - Error out with partial credentials error.
8. `ProfileA` contains `source_profile` set to `ProfileB` and static creds. `ProfileB` contains static creds.
    - Use static creds in `ProfileB` to call the sts assumeRole and return the credentials from the sts assumeRole call.


source/credentials_provider_profile.c
```diff
@@ -374,13 +384,30 @@ struct aws_credentials_provider *aws_credentials_provider_new_profile(
     struct aws_string *credentials_file_path = NULL;
     struct aws_string *config_file_path = NULL;
     struct aws_string *profile_name = NULL;
+    bool first_profile_in_chain = false;
+    if (source_profiles_table == NULL) {
+        source_profiles_table = aws_mem_calloc(allocator, 1, sizeof(struct aws_hash_table));
```
DmitriyMusatkin, line 389: hash table where each element is another hash table?

source/credentials_provider_profile.c
```diff
@@ -374,13 +384,30 @@ struct aws_credentials_provider *aws_credentials_provider_new_profile(
...
+        first_profile_in_chain = true;
+        if (aws_hash_table_init(
+                source_profiles_table,
+                allocator,
+                3,
+                aws_hash_byte_cursor_ptr,
+                (aws_hash_callback_eq_fn *)aws_byte_cursor_eq,
```
DmitriyMusatkin, line 396: any function type casting is ub. it will probably work in most cases, but its still better to wrap

source/credentials_provider_profile.c
```diff
@@ -428,11 +455,35 @@ struct aws_credentials_provider *aws_credentials_provider_new_profile(
         goto on_finished;
     }
     const struct aws_profile_property *role_arn_property = aws_profile_get_property(profile, s_role_arn_name);
+    bool profile_contains_access_key = aws_profile_get_property(profile, s_access_key_id_profile_var);
+    bool profile_contains_secret_access_key = aws_profile_get_property(profile, s_secret_access_key_profile_var);
+    bool profile_contains_credentials = profile_contains_access_key || profile_contains_secret_access_key;
```
DmitriyMusatkin, line 460: should this be end? cred should not be valid if it only contains one of the 2

waahm7: It's an `Or` condition so that we don't treat the profile as an STS profile, even if partial creds are present at https://github.com/awslabs/aws-c-auth/pull/224/files#diff-f7cca12a8538a30712ad4b18e39bb61a240457b820982e916f6f27e76628c64fR477. Later, we fail the creation as invalid if either of them is missing at https://github.com/awslabs/aws-c-auth/pull/224/files#diff-f7cca12a8538a30712ad4b18e39bb61a240457b820982e916f6f27e76628c64fR481. This handles cases where the config file is as follows:

```

[profile userA]
role_arn =  arn
source_profile = userB

[profile userB]
role_arn = arn
aws_access_key_id = accessJey
source_profile = userC

[profile userC]
aws_access_key_id = acess_key
secret_access_key_id = secret
```
We should fail the request at profileB due to partial credentials instead of using creds from profileC. I copied this behavior from the AWS CLI.

source/credentials_provider_profile.c
```diff
@@ -428,11 +456,35 @@ struct aws_credentials_provider *aws_credentials_provider_new_profile(
         goto on_finished;
     }
     const struct aws_profile_property *role_arn_property = aws_profile_get_property(profile, s_role_arn_name);
+    bool profile_contains_access_key = aws_profile_get_property(profile, s_access_key_id_profile_var);
+    bool profile_contains_secret_access_key = aws_profile_get_property(profile, s_secret_access_key_profile_var);
```
graebm, line 454: trivial:
```suggestion
    bool profile_contains_access_key = aws_profile_get_property(profile, s_access_key_id_profile_var) != NULL;
    bool profile_contains_secret_access_key = aws_profile_get_property(profile, s_secret_access_key_profile_var) != NULL;
```

source/credentials_provider_profile.c
```diff
@@ -374,6 +382,23 @@ struct aws_credentials_provider *aws_credentials_provider_new_profile(
...
+    if (source_profiles_table == NULL) {
+        source_profiles_table = aws_mem_calloc(allocator, 1, sizeof(struct aws_hash_table));
+        first_profile_in_chain = true;
+        /* source_profiles_table is an hashtable of (aws_byte_cursor *) -> NULL to detect recursion loop */
+        if (aws_hash_table_init(
+                source_profiles_table,
+                allocator,
+                3,
+                aws_hash_c_string,
+                aws_hash_callback_c_str_eq,
```
graebm: comment says it's `(aws_byte_cursor *) -> NULL`. But based on hashing functions, seems like we're doing `(const char *) -> NULL`

graebm (approved):

DmitriyMusatkin (approved):

PR #230, opened 2024-02-14
improve error handling

TingDaoK:
- raise proper error if error raise is missing
- Now the allocation can't fail, don't handle it.

### DIFF
--- a/source/credentials_provider_anonymous.c
+++ b/source/credentials_provider_anonymous.c
@@ -42,9 +42,6 @@ struct aws_credentials_provider *aws_credentials_provider_new_anonymous(
     struct aws_credentials_provider *provider = aws_mem_calloc(allocator, 1, sizeof(struct aws_credentials_provider));
 
     struct aws_credentials *credentials = aws_credentials_new_anonymous(allocator);
-    if (credentials == NULL) {
-        goto on_new_credentials_failure;
-    }
 
     aws_credentials_provider_init_base(provider, allocator, &s_aws_credentials_provider_anonymous_vtable, credentials);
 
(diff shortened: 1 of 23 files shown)


source/credentials_utils.c
```diff
@@ -275,6 +275,7 @@ struct aws_credentials *aws_parse_credentials_from_json_document(
     struct aws_json_value *document_root = aws_json_value_new_from_string(allocator, document);
     if (document_root == NULL) {
         AWS_LOGF_ERROR(AWS_LS_AUTH_CREDENTIALS_PROVIDER, "Failed to parse document as Json document.");
+        aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
```
graebm, line 278: we should fix this is json.c
aws_json_value_new_from_string() ought to raise an error
and that error should be AWS_ERROR_INVALID_JSON, which is more useful than INVALID_ARGUMENT

graebm (approved):

source/credentials_provider_anonymous.c
```diff
@@ -42,9 +42,6 @@ struct aws_credentials_provider *aws_credentials_provider_new_anonymous(
     struct aws_credentials_provider *provider = aws_mem_calloc(allocator, 1, sizeof(struct aws_credentials_provider));
 
     struct aws_credentials *credentials = aws_credentials_new_anonymous(allocator);
```
waahm7, line 44: Probably fine here, but I think it would be better to keep the null check as a design pattern. What if this function starts failing in the future? We should not make assumptions about the return value at this level.

waahm7 (approved):

PR #233, opened 2024-04-17
Update ECS Credentials Provider to add support for AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE

waahm7:
*Issue #, if available:*

*Description of changes:*
Adds support for AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE with the following priority
1. auth_token parameter
2. AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE (env var which contains absolute path to the token file. The file will be re-read for each call to get credentials.)
3. AWS_CONTAINER_AUTHORIZATION_TOKEN (env var which contains static auth token)


include/aws/auth/auth.h
```diff
@@ -51,6 +51,7 @@ enum aws_auth_errors {
     AWS_AUTH_CREDENTIALS_PROVIDER_SSO_SOURCE_FAILURE,
     AWS_AUTH_IMDS_CLIENT_SOURCE_FAILURE,
     AWS_AUTH_PROFILE_STS_CREDENTIALS_PROVIDER_CYCLE_FAILURE,
+    AWS_AUTH_CREDENTIALS_PROVIDER_ECS_INVALID_TOKEN_FILE_PATH,
```
DmitriyMusatkin, line 54: does sep say it should fail or just be treated as if it was not set?

source/credentials_provider_ecs.c
```diff
@@ -98,12 +103,33 @@ static struct aws_credentials_provider_ecs_user_data *s_aws_credentials_provider
         goto on_error;
     }
 
-    return wrapped_user_data;
+    struct aws_credentials_provider_ecs_impl *impl = ecs_provider->impl;
+    if (impl->auth_token_file_path) {
```
DmitriyMusatkin, line 107: nit: lets be consistent with how null checks are done? i've seen full != NULL in other places

source/credentials_provider_ecs.c
```diff
@@ -98,12 +103,33 @@ static struct aws_credentials_provider_ecs_user_data *s_aws_credentials_provider
...
+                &wrapped_user_data->auth_token,
+                ecs_provider->allocator,
+                aws_string_c_str(impl->auth_token_file_path))) {
+            AWS_LOGF_ERROR(
+                AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+                "(id=%p) ECS credentials provider failed to read token from the path: %s with error: %d",
+                (void *)ecs_provider,
+                aws_string_c_str(impl->auth_token_file_path),
+                aws_last_error());
+            aws_raise_error(AWS_AUTH_CREDENTIALS_PROVIDER_ECS_INVALID_TOKEN_FILE_PATH);
```
DmitriyMusatkin, line 118: should this fail or should this fallback to other methods?

include/aws/auth/auth.h
waahm7: Yes, it says If AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE is set, and the file cannot be opened or read for any reason (doesn’t exist, no permissions, invalid bytes, etc.), implementations must fail to resolve.

DmitriyMusatkin (approved):

PR #234, opened 2024-04-19
Enforce ECSCredentialsProvider IP Rules

waahm7:
*Description of changes:*
- The resolved ip address of ECSCredentials provider must follow certain limitations if it is HTTP. This PR enforces those limitations so that we don't make requests on potential malicious HTTP servers.


tests/credentials_provider_ecs_tests.c
```diff
@@ -536,6 +557,87 @@ static int s_credentials_provider_ecs_basic_success(struct aws_allocator *alloca
 
 AWS_TEST_CASE(credentials_provider_ecs_basic_success, s_credentials_provider_ecs_basic_success);
 
+static int s_credentials_provider_ecs_mocked_server_basic_ipv4_invalid(struct aws_allocator *allocator, void *ctx) {
```
graebm, line 560: There's already a system for mocking HTTP calls in auth tests. Is there a really good reason for adding another?

source/credentials_provider_ecs.c
```diff
@@ -430,9 +432,58 @@ static void s_ecs_query_task_role_credentials(struct aws_credentials_provider_ec
...
+    struct aws_credentials_provider_ecs_impl *impl = ecs_user_data->ecs_provider->impl;
+
+    if (impl->is_https) {
+        return true;
+    }
+
+    bool result = false;
+
+    const struct aws_byte_cursor address =
+        aws_byte_cursor_from_c_str(aws_http_connection_get_remote_endpoint(connection)->address);
```
graebm, line 453: This technique won't work with many other HTTP clients. It's uncommon that you have a chance to inspect the IP before making the actual request.

just asking the host_resolver directly would be more portable

I get that you're trying to avoid additional calls to the resolver. You could do like the Rust SDK and [only validate](https://github.com/awslabs/aws-sdk-rust/blob/29883b151225f7b27a22e4d343585d2dc8207521/sdk/aws-config/src/ecs.rs#L176-L178) when full-uri is passed in by the user.

Is the fact that this relies on the HTTP client what's forcing you to add a 2nd way of mocking HTTP calls?

source/credentials_provider_ecs.c
```diff
@@ -430,9 +432,58 @@ static void s_ecs_query_task_role_credentials(struct aws_credentials_provider_ec
...
+    if (aws_host_utils_is_ipv4(address)) {
+        const struct aws_byte_cursor ipv4_loopback_address_prefix = aws_byte_cursor_from_c_str("127.");
+        const struct aws_byte_cursor ecs_container_host_address = aws_byte_cursor_from_c_str("169.254.170.2");
+        const struct aws_byte_cursor eks_container_host_address = aws_byte_cursor_from_c_str("169.254.170.23");
+
+        result |= aws_byte_cursor_starts_with(&address, &ipv4_loopback_address_prefix);
+        result |= aws_byte_cursor_eq(&address, &ecs_container_host_address);
+        result |= aws_byte_cursor_eq(&address, &eks_container_host_address);
+    } else if (aws_host_utils_is_ipv6(address, false)) {
+        /* Check for both the short form and long form of an IPv6 address to be safe. */
```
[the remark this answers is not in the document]
waahm7: We are getting these IP strings from a system call, and we can convert them back to byte values using an OS dependent system call and memcmp them (at least that's what ChatGPT told me). I can add a function to common, but the fact that we are getting these IPs from system calls, we discussed that it should be good enough to just compare both the short form and the long form and ignore the weird ones.  

Testing the OS behavior for n different platforms led to the addition of an actual mock server, but unfortunately, GitHub runners don't support IPv6.

waahm7:
Closing in favor of #238

PR #236, opened 2024-05-02
Refactor ECSCredentialsProvider Part 1

waahm7:
*Description of changes:*
This PR refactors our ECSCredentialsProvider to have separate environment-specific credentials providers, making it easier to use in environments other than the default chain. This PR does not modify the STS CredentialsProvider to address the original issue that it does not support the ECSCredentialsProvider; that will be addressed in Part 2.


include/aws/auth/credentials.h
```diff
@@ -1009,6 +1024,20 @@ struct aws_credentials_provider *aws_credentials_provider_new_ecs(
     struct aws_allocator *allocator,
     const struct aws_credentials_provider_ecs_options *options);
 
+/**
+ * Creates a provider that sources credentials from the ecs role credentials service and reads the required params from
+ * environment variables
```
graebm, line 1029: tweak comments on aws_credentials_provider_new_ecs() to mention the alternative environment function. yes, this is also mentioned in the options-struct, but it doesn't hurt to be repetitive with important information like that

include/aws/auth/credentials.h
```diff
@@ -290,6 +276,35 @@ struct aws_credentials_provider_ecs_options {
...
+ * (http://169.254.170.2) or via a full URI specified by environment variables (in order of priority):
+ * AWS_CONTAINER_CREDENTIALS_RELATIVE_URI
+ * AWS_CONTAINER_CREDENTIALS_FULL_URI
+ *
+ * For the Authorization token, there are two ways (in order of priority):
+ * 1. AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE (an env var which contains the absolute path to the token file. The file
+ * will be re-read for each call to get credentials.)
+ * 2. AWS_CONTAINER_AUTHORIZATION_TOKEN (an env var that contains a static auth token)
+ */
+struct aws_credentials_provider_ecs_environment_options {
```
graebm, line 291: trivial/debatable: declare the environment-options struct before the explicit-options struct, so that people encounter it first when searching, since it's the more common use case

likewise, declare the new_ecs_from_environment() function before the new_ecs() function

source/credentials_provider_ecs.c
```diff
@@ -629,3 +628,93 @@ struct aws_credentials_provider *aws_credentials_provider_new_ecs(
...
+        explicit_options.port = aws_uri_port(&full_uri);
+        if (explicit_options.port == 0) {
+            explicit_options.port = explicit_options.tls_ctx ? 443 : 80;
+        }
+
+        provider = aws_credentials_provider_new_ecs(allocator, &explicit_options);
+
+    } else {
+        /* Neither environment variable is set */
+        aws_raise_error(AWS_AUTH_CREDENTIALS_PROVIDER_INVALID_ENVIRONMENT);
```
graebm, line 708: Current string for this error is [Valid credentials could not be sourced from process environment](https://github.com/awslabs/aws-c-auth/blob/277c8fbcb3f5341adb51000624b55ace301cfe82/source/auth.c#L42-L43)

Either make this string more generic, or add a more specific error-code

source/credentials_provider_ecs.c
```diff
@@ -629,3 +628,93 @@ struct aws_credentials_provider *aws_credentials_provider_new_ecs(
...
+    struct aws_string *full_uri_str = NULL;
+    struct aws_uri full_uri;
+    AWS_ZERO_STRUCT(full_uri);
+    struct aws_credentials_provider *provider = NULL;
+
+    /* Read the auth token variables */
+    struct aws_string *ecs_env_token_file_path = NULL;
+    struct aws_string *ecs_env_token = NULL;
+    if (aws_get_environment_value(allocator, s_ecs_creds_env_token_file, &ecs_env_token_file_path) != AWS_OP_SUCCESS ||
+        aws_get_environment_value(allocator, s_ecs_creds_env_token, &ecs_env_token) != AWS_OP_SUCCESS) {
```
graebm, line 652: trivial/annoying: `ecs_env_token_file_path` would leak if the 1st call was AWS_OP_SUCCESS, but the 2nd call was AWS_OP_ERR

But the funny thing is `aws_get_environment_value()` can't possibly fail (now that allocations can't fail)

Either properly handle each individual error, or just call aws_get_environment_value() as if it didn't return an error code

source/credentials_provider_ecs.c
```diff
@@ -629,3 +628,93 @@ struct aws_credentials_provider *aws_credentials_provider_new_ecs(
...
+        goto cleanup;
+    }
+    if (ecs_env_token_file_path != NULL) {
+        explicit_options.auth_token_file_path = aws_byte_cursor_from_string(ecs_env_token_file_path);
+    }
+    if (ecs_env_token != NULL) {
+        explicit_options.auth_token = aws_byte_cursor_from_string(ecs_env_token);
+    }
+
+    if (aws_get_environment_value(allocator, s_ecs_creds_env_relative_uri, &relative_uri_str) == AWS_OP_SUCCESS &&
```
graebm, line 662: trivial: if you chose not to check this error above, then also don't check it here. Maybe just put all 4 read-env-var calls in a row

source/credentials_provider_ecs.c
```diff
@@ -629,3 +628,93 @@ struct aws_credentials_provider *aws_credentials_provider_new_ecs(
...
+    } else if (
+        aws_get_environment_value(allocator, s_ecs_creds_env_full_uri, &full_uri_str) == AWS_OP_SUCCESS &&
+        full_uri_str != NULL && full_uri_str->len != 0) {
+
+        /* Using FULL_URI */
+        AWS_LOGF_INFO(
+            AWS_LS_AUTH_CREDENTIALS_PROVIDER, "ECS provider: using full uri %s", aws_string_c_str(full_uri_str));
+        struct aws_byte_cursor full_uri_cursor = aws_byte_cursor_from_string(full_uri_str);
+        if (aws_uri_init_parse(&full_uri, allocator, &full_uri_cursor)) {
+            goto cleanup;
```
graebm, line 686: Log about how AWS_CONTAINER_CREDENTIALS_FULL_URI was invalid. Someone hitting this error and relying on the default chain would have a hell of a time figuring that out

graebm: debatable: aws_uri_init_parse() failure raises `AWS_ERROR_MALFORMED_INPUT_STRING`

would this be better as AWS_AUTH_CREDENTIALS_PROVIDER_INVALID_ENVIRONMENT? something else?

source/credentials_provider_ecs.c
graebm: trivial: Also, LOGF about this, at DEBUG or TRACE level. We don't need it polluting logs when it happens from default credentials provider, but if someone was calling it directly, it would be nice to see **something**

source/credentials_provider_ecs.c
```diff
@@ -629,3 +628,93 @@ struct aws_credentials_provider *aws_credentials_provider_new_ecs(
 
     return NULL;
 }
+
+struct aws_credentials_provider *aws_credentials_provider_new_ecs_from_environment(
+    struct aws_allocator *allocator,
+    const struct aws_credentials_provider_ecs_environment_options *options) {
+
```
graebm, line 635: aws_raise_error(AWS_ERROR_INVALID_ARGUMENT) if anything is missing from options-struct. tls_ctx is the specifically the one I'd be worried about someone accidentally omitting.

Should we **require** tls_ctx? If it's optional, check later after reading the FULL_URI, in case it's "https://". Maybe simpler to just make it required. So someone doesn't build a program that **seems** to work until it's in an environment where FULL_URI is "https://"

include/aws/auth/credentials.h
waahm7: Thanks, updated.

source/credentials_provider_ecs.c
waahm7: I have made the AWS_AUTH_CREDENTIALS_PROVIDER_INVALID_ENVIRONMENT error more generic. I have already added a log in part 2, so not adding it here.

source/credentials_provider_ecs.c
waahm7: Updated the code ignore the return value from aws_get_environment_value as it makes the code simpler.

source/credentials_provider_ecs.c
waahm7: I have added a log, I think `AWS_ERROR_MALFORMED_INPUT_STRING` is fine since this will only be raised if someone has an invalid uri.

source/credentials_provider_ecs.c
waahm7: Thanks, I have made it a required parameter. We always have the tls_ctx whenever we need this API; so it's simpler to just always require it.

graebm (approved):

PR #239, opened 2024-05-10
Fix STSCredentialsProvider Signing Region

waahm7:
*Description of changes:*
- In #237, I forgot to update the region for signing config which resulted in errors if anyone tried to use a regional sts_endpoint. This PR fixes the signing config to use proper resolved region.

### DIFF
--- a/source/credentials_provider_sts.c
+++ b/source/credentials_provider_sts.c
@@ -48,7 +48,6 @@ static struct aws_http_header s_content_type_header = {
 
 static struct aws_byte_cursor s_content_length = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("content-length");
 static struct aws_byte_cursor s_path = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("/");
-static struct aws_byte_cursor s_signing_region = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("us-east-1");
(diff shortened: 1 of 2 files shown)


source/credentials_provider_sts.c
```diff
@@ -533,8 +533,13 @@ static void s_start_make_request(
     provider_user_data->signing_config.signed_body_header = AWS_SBHT_NONE;
     provider_user_data->signing_config.config_type = AWS_SIGNING_CONFIG_AWS;
     provider_user_data->signing_config.credentials_provider = impl->provider;
-    aws_date_time_init_now(&provider_user_data->signing_config.date);
-    provider_user_data->signing_config.region = s_signing_region;
+    uint64_t now = UINT64_MAX;
+    if (impl->system_clock_fn(&now) != AWS_OP_SUCCESS) {
```
DmitriyMusatkin, line 537: is this expected change?

waahm7: Are you asking about using the mocked system clock? We can avoid making this change and just assert on the prefix of the signature in tests; however, since we already had the option, it seemed more sensible to just use the mocked clock if set to keep it consistent.

DmitriyMusatkin: no was just checking whether you were intending to make this change or its something that slipped in accidentally. cause it looked somewhat unrelated to your changes. i see its for mocking now

DmitriyMusatkin (approved):

PR #244, opened 2024-07-29
Support "external_id" in config file, for STS AssumeRole

graebm:
Support reading "external_id" from the `config` file, similar to how [AWS CLI does it](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html#cli-configure-role-xaccount). Pass this value as "ExternalId" to [STS AssumeRole](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html#API_AssumeRole_RequestParameters).


source/credentials_provider_sts.c
```diff
@@ -780,29 +794,28 @@ struct aws_credentials_provider *aws_credentials_provider_new_sts(
 
     impl->role_session_name =
         aws_string_new_from_array(allocator, options->session_name.ptr, options->session_name.len);
-
-    if (!impl->role_session_name) {
-        goto on_done;
-    }
-
```
waahm7, line 787: We should throw a proper error if these strings are required instead of creating a string with 0 len.

waahm7 (approved):

PR #259, opened 2025-01-08
Fix Retries for ECSCredentialsProvider

waahm7:
*Description of changes:*
- Add retries to ECSCredentialsProvider

### DIFF
--- a/source/credentials_provider_ecs.c
+++ b/source/credentials_provider_ecs.c
@@ -165,16 +165,6 @@ on_error:
     return NULL;
 }
 
-static void s_aws_credentials_provider_ecs_user_data_reset_response(
-    struct aws_credentials_provider_ecs_user_data *ecs_user_data) {
-    ecs_user_data->current_result.len = 0;
-    ecs_user_data->status_code = 0;
-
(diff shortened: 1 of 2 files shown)


source/credentials_provider_ecs.c
```diff
@@ -71,6 +75,25 @@ struct aws_credentials_provider_ecs_user_data {
     int error_code;
 };
 
+/* called in between retries. */
+static void s_ecs_user_data_reset_request_specific_data(struct aws_credentials_provider_ecs_user_data *user_data) {
```
graebm, line 79: There's another similarly named pre-existing function in this file: `s_aws_credentials_provider_ecs_user_data_reset_response()`

What's the difference? Can they be combined? If not, put them next to each other with a comment explaining when you'd call one and not the other. 

If there's already some kind of reset function in this file ... did we miss some other flow that should be reporting record_success() (or failure)?

waahm7: Thanks, I have removed the other function since it was not needed. We will call the new function between retries to clear the request level data.

graebm (approved):

PR #260, opened 2025-01-13
Add account id

waahm7:
*Issue #, if available:*
https://github.com/awslabs/aws-crt-swift/issues/303

*Description of changes:*
- Adds a new optional account_id field to credentials. Currently, only static credentials support it, but soon we will add account_id for SSO, STS, and any other providers that can provide it.

### DIFF
--- a/include/aws/auth/credentials.h
+++ b/include/aws/auth/credentials.h
@@ -82,6 +82,7 @@ struct aws_credentials_provider_static_options {
     struct aws_byte_cursor access_key_id;
     struct aws_byte_cursor secret_access_key;
     struct aws_byte_cursor session_token;
+    struct aws_byte_cursor account_id;
 };
 
 /**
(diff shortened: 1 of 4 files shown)


.github/workflows/codecov.yml
```diff
@@ -28,4 +28,4 @@ jobs:
       run: |
         python3 -c "from urllib.request import urlretrieve; urlretrieve('${{ env.BUILDER_HOST }}/${{ env.BUILDER_SOURCE }}/${{ env.BUILDER_VERSION }}/builder.pyz?run=${{ env.RUN }}', 'builder')"
         chmod a+x builder
-        ./builder build -p ${{ env.PACKAGE_NAME }} --compiler=gcc-9 --cmake-extra=-DASSERT_LOCK_HELD=ON --coverage
+        ./builder build -p ${{ env.PACKAGE_NAME }} --compiler=gcc-12 --cmake-extra=-DASSERT_LOCK_HELD=ON --coverage
```
TingDaoK, line 31: why? Add some comments about this?
Do we need it to be applied to other codecov jobs?

waahm7: I have added the comment. Yes, we need to fix it everywhere where we updated the code coverage action ubuntu version.

TingDaoK: hmmm looks like the ci before the commit changing the compiler still work? https://github.com/awslabs/aws-c-auth/actions/runs/12757805423/job/35558824650

waahm7: If you expand the build step and scroll to the end, you can see the errors. No idea why the overall CI says success.

TingDaoK: hmmm, maybe instead of updating the gcc version for every codecov ci, we should update our builder

https://github.com/awslabs/aws-crt-builder/blob/42d15ba4ee3fbe0e812e25094a8cf59eaef0cf1e/builder/actions/cmake.py#L254-L258

We install our own version of gcc, but the `ctest -T coverage` probably still use the default one in the system.

Seems like we can pass `--gcov` to chose what version of gcov to solve the version mismatch thing. https://stackoverflow.com/questions/12454175/gcov-out-of-memory-mismatched-version

waahm7: Thanks, fixed by https://github.com/awslabs/aws-crt-builder/pull/317.

TingDaoK (approved):

PR #262, opened 2025-01-29
Account ID support for Crendentials Providers

waahm7:
Issue: https://github.com/awslabs/aws-crt-swift/issues/303

*Description of changes:*
Update the following providers to provide account_id if present. 

1. **EnvironmentCredentialsProvider**: Checks `AWS_ACCOUNT_ID` env variable.
2. **SSOCredentialsProvider:** Account id is parsed from the `sso_account_id` config parameter which was also required for the request.
3. **STS:** Account_id is parsed from the Arn in the response. 
4. **STSWebIdentity:** Account_id is parsed from the Arn in the response. 
5.  **ProcessCredentialsProvider:** Account_id is parsed from `AccountId` field in the response if present.
6. **ECSProvider:** Account_id is parsed from `AccountId` field in the response if present.
7. **ProfileProvider:** Parses `aws_account_id` in the config file.


include/aws/auth/credentials.h
```diff
@@ -797,6 +797,28 @@ struct aws_credentials *aws_credentials_new_from_string(
...
+ * @return a valid credentials object, or NULL
+ */
+AWS_AUTH_API
+struct aws_credentials *aws_credentials_new_from_string_with_account_id(
+    struct aws_allocator *allocator,
+    const struct aws_string *access_key_id,
+    const struct aws_string *secret_access_key,
+    const struct aws_string *session_token,
+    const struct aws_string *account_id,
+    uint64_t expiration_timepoint_seconds);
```
graebm, line 820: There's a time-honored tradition in this codebase, when new arguments keep creeping into constructors...
```suggestion
struct aws_credentials *aws_credentials_new_from_options(
    struct aws_allocator *allocator,
    const struct aws_credentials_options *options);
```

source/credentials_provider_sts.c
```diff
@@ -229,6 +234,40 @@ static int s_sts_xml_on_AssumeRoleResult_child(struct aws_xml_node *node, void *
...
+    if (aws_byte_cursor_eq_c_str_ignore_case(&node_name, "Arn")) {
+        if (aws_xml_node_as_body(node, &arn_cursor)) {
+            return AWS_OP_ERR;
+        }
+        struct aws_byte_cursor account_id;
+        AWS_ZERO_STRUCT(account_id);
+        /* The format of the Arn is arn:partition:service:region:account-id:resource-ID and we need to parse the
+         * account-id out of it which is the fifth element. */
+        for (int i = 0; i < 5; i++) {
+            if (!aws_byte_cursor_next_split(&arn_cursor, ':', &account_id)) {
```
DmitriyMusatkin, line 259: note: its valid for arn not to have account id field. most fields are optional from what i remember (s3 is an example of arns without a bunch of fields). do we want to have logic that accounts for that? or just setting it to empty string is fine?

source/credentials_provider_sts_web_identity.c
```diff
@@ -261,15 +266,45 @@ static int s_stswebid_200_xml_on_AssumeRoleWithWebIdentityResponse_child(
...
+    return AWS_OP_SUCCESS;
+}
+
+static int s_stswebid_xml_on_AssumedRoleUser_child(struct aws_xml_node *node, void *user_data) {
+    struct sts_web_identity_user_data *query_user_data = user_data;
+    struct aws_byte_cursor node_name = aws_xml_node_get_name(node);
+    struct aws_byte_cursor arn_cursor;
+    AWS_ZERO_STRUCT(arn_cursor);
+
+    if (aws_byte_cursor_eq_c_str_ignore_case(&node_name, "Arn")) {
```
DmitriyMusatkin, line 287: maybe pull arn parsing into a separate helper that we can test standalone?

source/credentials_utils.c
```diff
@@ -189,26 +190,39 @@ struct aws_credentials *aws_parse_credentials_from_aws_json_object(
...
         goto done;
     }
 
     struct aws_byte_cursor token_cursor;
+    AWS_ZERO_STRUCT(token_cursor);
     if (options->token_name) {
         token = aws_json_value_get_from_object(document_root, aws_byte_cursor_from_c_str((char *)options->token_name));
-        if (!aws_json_value_is_string(token) || aws_json_value_get_string(token, &token_cursor) == AWS_OP_ERR) {
+        if (!aws_json_value_is_string(token) || aws_json_value_get_string(token, &token_cursor) == AWS_OP_ERR ||
+            token_cursor.len == 0) {
```
graebm, line 207: why is this one getting a length check, but `access_key_id`  and `secret_access_key` above don't get a length check?

worth breaking these 4 checks out into a helper function?

include/aws/auth/credentials.h
waahm7: There are multiple types of credentials and they have new functions like `new_ecs`, `new_anonymous`, `new`, and `new_from_string`. Internally, they're represented with a union https://github.com/awslabs/aws-c-auth/blob/main/source/credentials.c#L80 and the actual credentials type is supposed to be opaque. I think consolidating everything into an option struct would be a mess. 

We can consolidate `new` and `new_with_account_id` function into an option struct and `new_from_string` and `new_from_string_with_account_id` into an option struct but we will potentially need new structs for new types.

source/credentials_provider_sts.c
waahm7: account_id is an optional nice-to-have field, so it's fine if it's blank.

source/credentials_utils.c
waahm7: `access_key_id` and `secret_access_key` are required parameters for credentials, so we will fail to create the credentials struct if they are empty. So we don't need to do any length check here. `aws_session_token` is an optional field, but this parser has an option of `token_required`, if that is set, we want to error out if it's empty.

include/aws/auth/credentials.h
waahm7: Thanks, updated with the new_from_options function.

include/aws/auth/credentials.h
```diff
@@ -758,13 +776,9 @@ struct aws_credentials *aws_credentials_new(
  * @return a valid credentials object, or NULL
  */
 AWS_AUTH_API
-struct aws_credentials *aws_credentials_new_with_account_id(
```
waahm7, line 761: I added the function recently, https://github.com/awslabs/aws-c-auth/pull/260 therefore it's safe to remove.

include/aws/auth/credentials.h
```diff
@@ -715,6 +715,24 @@ struct aws_credentials_provider_cognito_options {
...
+
+    /* Value for the aws access key id field */
+    struct aws_byte_cursor access_key_id_cursor;
+    /* Value for the secret access key field */
+    struct aws_byte_cursor secret_access_key_cursor;
+    /* (Optional) security token associated with the credentials */
+    struct aws_byte_cursor session_token_cursor;
+    /* (Optional) account id associated with the credentials */
+    struct aws_byte_cursor account_id_cursor;
+    /* expiration_timepoint_seconds timepoint, in seconds since epoch, that the credentials will no longer be valid.
```
graebm, line 731: trivial: none of the other docs copy/pasted the variable name
```suggestion
    /* Time point, in seconds since epoch, that the credentials will no longer be valid.
```

graebm (approved):

include/aws/auth/credentials.h
graebm: sorry I'm annoying I know

Issue #257, opened 2024-12-30
Support for AWS_ENDPOINT_URL_STS


pawalt:
### Describe the feature

We'd like to provide `AWS_ENDPOINT_URL_STS` in order to point `aws-c-auth` at a local STS provider. Currently, this library only supports `sts.amazonaws.com`.

### Use Case

We'd like to use [mountpoint-s3](https://github.com/awslabs/mountpoint-s3) for to mount buckets using OIDC authentication. This works fine against production AWS, but we want to use a [MinIO](https://min.io/docs/minio/linux/operations/external-iam/configure-openid-external-identity-management.html) server in development so we can write automated tests. For this, we need a custom STS URL so `mountpoint-s3` can dial into MinIO instead of AWS.

### Proposed Solution

Allow the default STS URL to be set:
https://github.com/awslabs/aws-c-auth/blob/b965ffe238d3ec2adbe088ac9e1e0f8c4126c436/source/credentials_provider_sts.c#L866

### Other Information

Crosslinking https://github.com/awslabs/mountpoint-s3/issues/1203

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

waahm7:
Hi, thank you for creating the issue. We are discussing the feasibility of this request. Could you please provide a bit more detail about why this is needed?

>  Unfortunately I am trying to test whether the Assume role provider works with mount-s3, so using the process credentials provider will defeat my purpose.

Default credentials provider is an implementation detail, and this sounds like trying to test the internal functionality of a dependency. What's the need for trying to test assumeRole? We have many unit tests that assert the implementation of assumeRole is working correctly.

waahm7:
@pawalt Sorry, I still don’t understand why something like `credential_process = aws sts assume-role-with-web-identity ....` won’t work. It will just use the CLI STSWebIdentity provider instead of the aws-c-auth one and mountpoint won't hit `sts.amazonaws.com`

waahm7:
Thank you for creating the issue. It is fixed in https://github.com/awslabs/aws-c-auth/releases/tag/v0.8.6 and now CRT will respect the environment variable.

pawalt:
@waahm7 thanks so much!!

PR #263, opened 2025-02-04
Support Endpoint Override for CredentialsProviders

waahm7:
*Issue #, if available:*
https://github.com/awslabs/aws-c-auth/issues/257
https://github.com/awslabs/aws-crt-swift/issues/311

*Description of changes:*
Support endpoint override for credentials providers.

The order of resolution for configured endpoint is as follows:
1. The value provided by a service-specific environment variable, `AWS_ENDPOINT_URL_<SERVICE>`. `<SERVICE>` here is the uppercased version of corresponding service identifier in [this official list](https://docs.aws.amazon.com/sdkref/latest/guide/ss-endpoints-table.html).
2. The value provided by the global endpoint environment variable, `AWS_ENDPOINT_URL`.
3. The value provided by a service-specific parameter from a services definition section referenced in a profile in the shared configuration file. The name of the service must match the identifier in [this official list](https://docs.aws.amazon.com/sdkref/latest/guide/ss-endpoints-table.html). E.g.:
```
[profile profile-with-services]
services = dev
endpoint_url = http://localhost:5567

[services dev]
sts = 
  endpoint_url = https://play.min.io:9000
```
4. The value provided by the global parameter from a profile in the shared configuration file. E.g.:
```
[profile profile-with-services]
services = dev
endpoint_url = http://localhost:5567
```


source/credentials_provider_sts.c
```diff
@@ -697,44 +699,66 @@ AWS_STATIC_STRING_FROM_LITERAL(s_region_config, "region");
...
-    struct aws_allocator *allocator,
-    const struct aws_credentials_provider_sts_options *options) {
+static struct aws_string *s_resolve_region(struct aws_allocator *allocator, const struct aws_profile *profile) {
     /* check environment variable first */
     struct aws_string *region = aws_credentials_provider_resolve_region_from_env(allocator);
     if (region != NULL && region->len > 0) {
         return region;
     }
 
+    if (profile) {
```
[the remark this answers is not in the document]
graebm: oh man, if we're adding new get-env functions, we can fix the terrible API of the current one (bad because it did out-values to handle reporting OOM which we don't do anymore, and bad because it didn't just take `char *` which forces some users to allocate/cleanup a needless aws_string)
```
struct aws_string *aws_getenv_nonempty(struct aws_allocator *, const char *); // NULL if missing or ""
struct aws_string *aws_getenv_raw(struct aws_allocator *, const char *); // may be ""
```

include/aws/auth/private/credentials_utils.h
```diff
@@ -166,15 +166,22 @@ AWS_AUTH_API
...
- * Constructs an endpoint in the format of service_name.region.amazonaws.com
+ * If an endpoint override is configured, use that. Otherwise do the following:
+ * Construct an endpoint in the format of service_name_host.region.amazonaws.com.
  * If the region is cn-north-1 or cn-northwest-1, .cn is appended to support China-specific regional endpoints.
+ * To configure the endpoint override, check the
+ * https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html.
  */
 AWS_AUTH_API
-int aws_credentials_provider_construct_regional_endpoint(
+int aws_credentials_provider_construct_endpoint(
```
graebm, line 176: I'm not proposing an options-struct because this is private

But if someone moves this out of private/ and is like "WHY SHOULD I CHANGE IT YOU ALREADY APPROVED THE EXISTING CODE"

then I can point at this comment here and say "NUH UH"

source/credentials_provider_sts.c
```diff
@@ -697,44 +699,66 @@ AWS_STATIC_STRING_FROM_LITERAL(s_region_config, "region");
...
+    if (aws_credentials_provider_construct_endpoint(
+            allocator,
+            out_endpoint,
+            *out_region,
+            s_sts_service_name,
+            s_sts_service_env_name,
+            s_sts_service_name,
+            profile_collection,
+            profile)) {
         goto cleanup;
```
graebm, line 756: if something goes wrong here, we end up setting `out_region` but not `out_endpoint`?

is that failure? or is that something we expect to happen?

source/credentials_provider_sts.c
```diff
@@ -697,44 +699,66 @@ AWS_STATIC_STRING_FROM_LITERAL(s_region_config, "region");
...
     }
-    const struct aws_profile *profile = aws_profile_collection_get_profile(profile_collection, profile_name);
-    if (!profile) {
+
+    *out_region = s_resolve_region(allocator, profile);
+
+    if (aws_credentials_provider_construct_endpoint(
+            allocator,
+            out_endpoint,
+            *out_region,
```
graebm, line 750: code clarity: `*out_region` could be NULL at this point....
and I see that `aws_credentials_provider_construct_endpoint()` only requires region *sometimes*

if region isn't always required, add a comment here to that effect

but if it is always required, maybe early-out if s_resolve_region() returns NULL

source/credentials_utils.c
```diff
@@ -383,23 +383,108 @@ static struct aws_byte_cursor s_aws_isoe_dns_suffix = AWS_BYTE_CUR_INIT_FROM_STR
...
+    }
+    service_endpoint_str = aws_string_new_from_buf(allocator, &service_endpoint_buf);
+
+    if (aws_get_environment_value(allocator, service_endpoint_str, &out_endpoint) == AWS_OP_SUCCESS &&
+        out_endpoint != NULL) {
+        goto on_finish;
+    }
+
+    /* Check AWS_ENDPOINT_URL variable */
+    if (aws_get_environment_value(allocator, s_endpoint_url_env, &out_endpoint) == AWS_OP_SUCCESS &&
```
graebm, line 421: copy/paste bug? setting `out_endpoint` twice, which will leak the value from the first call

source/credentials_utils.c
```diff
@@ -383,23 +383,108 @@ static struct aws_byte_cursor s_aws_isoe_dns_suffix = AWS_BYTE_CUR_INIT_FROM_STR
...
+    if (property) {
+        out_endpoint = aws_string_new_from_string(allocator, aws_profile_property_get_value(property));
+
+        goto on_finish;
+    }
+
+on_finish:
+    aws_byte_buf_clean_up(&service_endpoint_buf);
+    aws_string_destroy(service_endpoint_str);
+    return out_endpoint;
```
graebm, line 458: there are a lot of paths where NULL is returned, but no aws_raise_error() happened. Is that OK for this helper function?

source/credentials_utils.c
graebm: oh nevermind if != NULL goto finish

source/credentials_utils.c
waahm7: Yes, NULL means that there was no endpoint_override configured.

source/credentials_provider_sts.c
waahm7: Logically, it won't happen since if we resolve the region, we can construct the endpoint. But even if it does happen in the future, we have a fallback: We will check that out_endpoint is not set and then use the global STS endpoint.

source/credentials_provider_sts.c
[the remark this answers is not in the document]
waahm7: Thanks, I have added the `aws_getenv_nonempty` which will resolve this issue.

DmitriyMusatkin (approved):

PR #276, opened 2025-11-05
create a common base for http client, migrate sso

sbiscigl:
*Description of changes:*

Refactors the sso credentials client to a common "http credentials provider" base. this base provider exposes several callbacks for creating implementation specific requests and data for the generalized http credentials provider to use.

### DIFF
--- a/include/aws/auth/private/aws_http_credentials_provider.h
+++ b/include/aws/auth/private/aws_http_credentials_provider.h
@@ -41,7 +41,7 @@ struct aws_credentials_provider_http_options {
      * Endpoint override for service endpoint. Leave null
      * to use default endpoint.
      */
-    struct aws_string *endpoint;
+    struct aws_byte_cursor endpoint;
 
     /*
      * Optional
@@ -77,15 +77,30 @@ struct aws_http_query_context {
     // implementation-specific data and errors
     void *parameters;
     void *request_data;
-    enum aws_auth_errors error;
 };
 
 typedef int(create_request_fn)(struct aws_http_query_context *query_context, void *user_data);
-typedef struct aws_credentials_options(
-    create_credentials_options_fn)(struct aws_credentials *credentials, struct aws_http_query_context *query_context);
+typedef void(finalize_credentials_fn)(struct aws_http_query_context *query_context);
 typedef void(clean_up_parameters_fn)(void *parameters);
-typedef void *(create_request_data_fn)(struct aws_allocator *allocator);
-typedef void(destroy_request_data_fn)(void *request_data);
+typedef void(create_request_data_fn)(struct aws_http_query_context *query_context);
+typedef void(reset_request_data_fn)(struct aws_http_query_context *query_context);
+
+/**
+ * Implementation specific userdata for the http provider.
+ */
+struct aws_http_credentials_provider_user_data {
+    /*
+     * owning pointer to implementation specific data. i.e. values that
+     * could be used during request creation.
+     */
+    void *parameters;
+
+    /*
+     * implementation specific callbacks for creating and cleaning up
+     * the underlying http request,
+     */
+    struct aws_http_credentials_provider_request_vtable *request_vtable;
+};
 
 /**
  * A table to hold implementation specific values for
@@ -101,7 +116,7 @@ struct aws_http_credentials_provider_request_vtable {
      * Creates credentials options. This is used to create the
      * credentials that are returned to the user
      */
-    create_credentials_options_fn *create_credentials_options_fn;
+    finalize_credentials_fn *finalize_credentials_fn;
 
     /*
      * cleans up the parameters that are provided to the implemetatnion.
@@ -116,13 +131,7 @@ struct aws_http_credentials_provider_request_vtable {
     /*
      * implementation specific data that is destroyed each request.
      */
-    destroy_request_data_fn *destroy_request_data_fn;
-
-    /*
-     * owning pointer to implementation specific data. i.e.e values that
-     * could be used during request creation.
-     */
-    void *parameters;
+    reset_request_data_fn *reset_request_data_fn;
 };
 
 AWS_EXTERN_C_BEGIN
@@ -134,7 +143,7 @@ AWS_EXTERN_C_BEGIN
  * @param provider allocated instance of provider that will be initialized with
  * a http based credentials fetching implementation.
  * @param options standard http options.
- * @param request_vtable implementation specific branching for http based providers
+ * @param user_data implementation specific branching for http based providers
  * such as hot to create requests.
  * @return error code if encountered.
  */
@@ -143,7 +152,7 @@ int aws_http_credentials_provider_init_base(
     struct aws_allocator *allocator,
     struct aws_credentials_provider *provider,
     struct aws_credentials_provider_http_options *options,
-    struct aws_http_credentials_provider_request_vtable *request_vtable);
+    struct aws_http_credentials_provider_user_data *user_data);
 
 AWS_EXTERN_C_END
 AWS_POP_SANE_WARNING_LEVEL
--- a/source/aws_http_credentials_provider.c
+++ b/source/aws_http_credentials_provider.c
@@ -24,15 +24,14 @@ struct aws_credentials_provider_http_impl {
     const struct aws_auth_http_system_vtable *function_table;
     struct aws_string *endpoint;
     struct aws_retry_strategy *retry_strategy;
-    struct aws_http_credentials_provider_request_vtable *request_vtable;
+    struct aws_http_credentials_provider_user_data *user_data;
 };
 
 static void s_http_query_context_reset_request_specific_data(struct aws_http_query_context *http_query_context) {
     struct aws_credentials_provider_http_impl *impl = http_query_context->provider->impl;
-    impl->request_vtable->destroy_request_data_fn(http_query_context->request_data);
+    impl->user_data->request_vtable->reset_request_data_fn(http_query_context);
     if (http_query_context->request) {
-        aws_http_message_release(http_query_context->request);
-        http_query_context->request = NULL;
+        http_query_context->request = aws_http_message_release(http_query_context->request);
     }
     if (http_query_context->connection) {
         int result = impl->function_table->aws_http_connection_manager_release_connection(
@@ -71,7 +70,7 @@ static struct aws_http_query_context *s_http_query_context_new(
     http_query_context->provider = aws_credentials_provider_acquire(provider);
     http_query_context->original_user_data = user_data;
     http_query_context->original_callback = callback;
-    http_query_context->parameters = impl->request_vtable->parameters;
+    http_query_context->parameters = impl->user_data->parameters;
     aws_byte_buf_init(&http_query_context->payload, provider->allocator, HTTP_RESPONSE_SIZE_INITIAL);
 
     return http_query_context;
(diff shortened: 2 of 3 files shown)


include/aws/auth/private/aws_http_credentials_provider.h
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+ * Configuration needed to create a underlying http client.
+ */
+struct aws_credentials_provider_http_options {
+    struct aws_credentials_provider_shutdown_options shutdown_options;
+    /*
+     * Connection bootstrap to use for any network connections
+     * made while sourcing credentials.
+     * Required.
```
TingDaoK: trivial: let's put required to the beginning of the doc. So, it's more clear which ones are required, and which ones are optional.

source/aws_http_credentials_provider.c
```diff
@@ -0,0 +1,612 @@
...
+        HTTP_CONNECT_TIMEOUT_DEFAULT_IN_SECONDS, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_MILLIS, NULL);
+
+    struct aws_http_connection_manager_options manager_options;
+    AWS_ZERO_STRUCT(manager_options);
+    manager_options.bootstrap = options->bootstrap;
+    manager_options.initial_window_size = HTTP_RESPONSE_SIZE_LIMIT;
+    manager_options.socket_options = &socket_options;
+    manager_options.host = host;
+    manager_options.port = 443;
+    manager_options.max_connections = 2;
```
TingDaoK: Maybe move this to the top as a defined const as well?

source/aws_http_credentials_provider.c
```diff
@@ -0,0 +1,612 @@
...
+
+    struct aws_credentials_provider_http_impl *impl =
+        aws_mem_acquire(allocator, sizeof(struct aws_credentials_provider_http_impl));
+
+    AWS_ZERO_STRUCT(*impl);
+    impl->request_vtable = request_vtable;
+
+    aws_credentials_provider_init_base(provider, allocator, &s_aws_credentials_provider_http_vtable, impl);
+
+    struct aws_tls_connection_options tls_connection_options;
```
TingDaoK:
```suggestion
    struct aws_tls_connection_options tls_connection_options;
    AWS_ZERO_STRUCT(tls_connection_options);
```
Otherwise, when we do `aws_tls_connection_options_clean_up` from `on_error` couple lines below, it will be garbage.

source/aws_http_credentials_provider.c
```diff
@@ -0,0 +1,612 @@
...
+    impl->function_table = options->function_table;
+    if (impl->function_table == NULL) {
+        impl->function_table = g_aws_credentials_provider_http_function_table;
+    }
+
+    impl->token_provider = aws_credentials_provider_acquire(options->token_provider);
+    impl->endpoint = aws_string_new_from_string(allocator, options->endpoint);
+    impl->account_id = aws_string_new_from_string(allocator, options->account_id);
+    impl->connection_manager = impl->function_table->aws_http_connection_manager_new(allocator, &manager_options);
+    if (impl->connection_manager == NULL) {
```
TingDaoK: trivial: add an error level log here as well.

source/aws_http_credentials_provider.c
```diff
@@ -0,0 +1,612 @@
...
+            "(id=%p): failed to create a retry strategy with error %s",
+            (void *)provider,
+            aws_error_debug_str(aws_last_error()));
+        goto on_error;
+    }
+
+    aws_tls_connection_options_clean_up(&tls_connection_options);
+    return AWS_OP_SUCCESS;
+on_error:
+    aws_credentials_provider_destroy(provider);
```
TingDaoK: hmmm, so when the initialization failed, we will destroy the provider here, which will actually free the memory for `provider` as well, according to https://github.com/awslabs/aws-c-auth/pull/276/files#diff-7df01304b24de64755d81ee19179ffcbb204080bfa33965734cf521989c455a5R476.

This seems like a concern, and may cause double free from https://github.com/awslabs/aws-c-auth/pull/276/files#diff-23acef6a19dca102219feb72adf38a73142c400f8870d91dfcde1f6a7a43cf64R286?

I'd suggest we only clean up the impl here, and leave the passed-in `provider` as it is, so that it's caller's responsibility to properly handle the clean up of the passed-in arg.

include/aws/auth/private/aws_http_credentials_provider.h
```diff
@@ -0,0 +1,145 @@
...
+    make_request_fn *make_request_fn;
+    /*
+     * Creates the headers that are send alongside
+     * the request to the underlying http client/
+     */
+    create_headers_fn *create_headers_fn;
+    /*
+     * Fetches the authentication token to be used alongside the requst.
+     */
+    credentials_get_token_fn *credentials_get_token_fn;
```
TingDaoK: just think out load, can we combine those three functions into one?
Let the specific provider to decide how to make the http request? 
If I understand this correctly, looks like for SSO provider, before we make the request, we ask for the token and add the token to the request as a header and then make the request.

Maybe give the specific provider the http request is going to make, then let it decide what to change the request to be like to fetch the credentials from the server?

sbiscigl: i tried to combine the functions, and it did make it a lot cleaner, i did have to add more callbacks to create/clean up data both at the client level and request level however.

include/aws/auth/private/aws_http_credentials_provider.h
```diff
@@ -0,0 +1,151 @@
...
+     * allowed to contain.
+     */
+    size_t max_connections;
+
+    /*
+     * Optional.
+     * Endpoint override for service endpoint. Leave null
+     * to use default endpoint.
+     */
+    struct aws_string *endpoint;
```
DmitriyMusatkin: who owns it? thruthfully im not a big fun of passing strings into options

source/aws_http_credentials_provider.c
```diff
@@ -0,0 +1,515 @@
...
+    impl->request_vtable->destroy_request_data_fn(http_query_context->request_data);
+    if (http_query_context->request) {
+        aws_http_message_release(http_query_context->request);
+        http_query_context->request = NULL;
+    }
+    if (http_query_context->connection) {
+        int result = impl->function_table->aws_http_connection_manager_release_connection(
+            impl->connection_manager, http_query_context->connection);
+        (void)result;
+        AWS_ASSERT(result == AWS_OP_SUCCESS);
```
DmitriyMusatkin: asserts are debug only and of rather limited use. (not sure we even run debug ci in this repo). 
is this an error that can happen in theory and is it catastrophic? can we bubble it up?

source/aws_http_credentials_provider.c
```diff
@@ -0,0 +1,515 @@
...
+            AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+            "(id=%p): failed to create a retry strategy with error %s",
+            (void *)provider,
+            aws_error_debug_str(aws_last_error()));
+        goto on_error;
+    }
+
+    aws_tls_connection_options_clean_up(&tls_connection_options);
+    return AWS_OP_SUCCESS;
+on_error:
```
DmitriyMusatkin: clean up impl you allocated?

include/aws/auth/private/aws_http_credentials_provider.h
```diff
@@ -0,0 +1,151 @@
...
+    struct aws_retry_token *retry_token;
+    struct aws_byte_buf path_and_query;
+
+    int status_code;
+    int error_code;
+
+    // implementation-specific data and errors
+    void *parameters;
+    void *request_data;
+    enum aws_auth_errors error;
```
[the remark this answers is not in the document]
TingDaoK: +1, looks like it only used [here](https://github.com/awslabs/aws-c-auth/pull/276/files#diff-23acef6a19dca102219feb72adf38a73142c400f8870d91dfcde1f6a7a43cf64R213).

Looks like we can just reuse the error_code.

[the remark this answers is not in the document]
sbiscigl: basically the reason i do this is because at the end of the current query for credentials we fail for a [implementation specific error code](https://github.com/awslabs/aws-c-auth/blob/main/source/credentials_provider_sso.c#L183-L185). I weighed out adding another "post response" callback or the specific error, and i went this way. Its more or less "I need to provide a custom error to return at the end"

source/aws_http_credentials_provider.c
```diff
@@ -0,0 +1,515 @@
...
+}
+
+static void s_http_query_context_destroy(struct aws_http_query_context *http_query_context) {
+    if (http_query_context == NULL) {
+        return;
+    }
+
+    s_http_query_context_reset_request_specific_data(http_query_context);
+    aws_mem_release(http_query_context->allocator, http_query_context->request_data);
+    aws_byte_buf_clean_up(&http_query_context->payload);
```
TingDaoK: We should move this into the `s_http_query_context_reset_request_specific_data`. On failure, there could also be some payload in the response received. And for the retry, we should have clean up the previous received payload as well.

source/aws_http_credentials_provider.c
```diff
@@ -0,0 +1,515 @@
...
+    aws_byte_buf_clean_up(&http_query_context->path_and_query);
+}
+
+static void s_http_query_context_destroy(struct aws_http_query_context *http_query_context) {
+    if (http_query_context == NULL) {
+        return;
+    }
+
+    s_http_query_context_reset_request_specific_data(http_query_context);
+    aws_mem_release(http_query_context->allocator, http_query_context->request_data);
```
TingDaoK: I think we should rely on `impl->request_vtable->destroy_request_data_fn` to release the memory, instead of adding an extra mem_release here.

source/aws_http_credentials_provider.c
```diff
@@ -0,0 +1,515 @@
...
+        http_query_context->error_code = AWS_AUTH_CREDENTIALS_PROVIDER_HTTP_STATUS_FAILURE;
+    }
+
+    /*
+     * If we can retry the request based on error response or http status code failure, retry it, otherwise, call the
+     * finalize function.
+     */
+    if (error_code || http_query_context->status_code != AWS_HTTP_STATUS_CODE_200_OK) {
+        enum aws_retry_error_type error_type =
+            aws_credentials_provider_compute_retry_error_type(http_query_context->status_code, error_code);
```
TingDaoK: it's not really part of the PR, but the status code can be 0, in case we don't have the full response back, in which case, we will rely on the error code itself to determine it's retriable or not. 

But the current `aws_credentials_provider_compute_retry_error_type`, doesn't seem to handle the 0 status code case well, it will just fallback to `AWS_RETRY_ERROR_TYPE_SERVER_ERROR`. 

Either we just default the status code to something like 400, or we handle the case where status code is not set in the helper.

source/aws_http_credentials_provider.c
sbiscigl: [not a great reason but this was how it was designed before](https://github.com/awslabs/aws-c-auth/blob/main/source/credentials_provider_sso.c#L71-L78). for the scope of this i dont want to change this behavior, but also would if we have a reason to change it right now.

source/aws_http_credentials_provider.c
```diff
@@ -0,0 +1,515 @@
...
+            impl->request_vtable->create_credentials_options_fn(credentials, http_query_context);
+        credentials_with_account_id = aws_credentials_new_with_options(http_query_context->allocator, &creds_option);
+    } else {
+        AWS_LOGF_ERROR(
+            AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+            "(id=%p) failed to query credentials",
+            (void *)http_query_context->provider);
+
+        if (http_query_context->error_code == AWS_ERROR_SUCCESS) {
+            http_query_context->error_code = http_query_context->error;
```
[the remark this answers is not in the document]
sbiscigl: [thats how it was before](https://github.com/awslabs/aws-c-auth/blob/main/source/credentials_provider_sso.c#L177-L186), i will change if it pasts tests, but as the running theme continues, scared of breaking backwards compat by making some of these changes, even through this one in particular doesnt make sense.

source/aws_http_credentials_provider.c
sbiscigl: > But the current aws_credentials_provider_compute_retry_error_type

I know this is not the best answer but i kind of want to avoid touching that for the sake of this PR because it will touch _all credentials provider_ and i want to reduce the radius of this change if possible. I am open to working this into this if you want it though

source/aws_http_credentials_provider.c
sbiscigl: this was how it was done before, and it breaks retry unit tests when i change this. As much as i want to change it, i dont want to battle the previous implementation

TingDaoK: looks like we are not addressing this and the following one?

PR #291, opened 2026-04-15
fix: Remove strict requirement for ECS SessionToken

blairdrummond:
I discovered that this library handles the `Token` field in the ECS authentication responses slightly differently than the other AWS SDKs do by checking that the Token is non-empty. The example JSON response in the code is this

```json
 {
  "Code" : "Success",
  "LastUpdated" : "2019-05-28T18:03:09Z",
  "Type" : "AWS-HMAC",
  "AccessKeyId" : "...",
  "SecretAccessKey" : "...",
  "Token" : "...",
  "Expiration" : "2019-05-29T00:21:43Z"
 }
```

However the `Token` / SessionToken is actually optional in the response with some systems where only the ACCESS_KEY and SECRET_KEY are used. By removing this constraint, the authentication works with more systems

### DIFF
--- a/source/credentials_provider_ecs.c
+++ b/source/credentials_provider_ecs.c
@@ -190,7 +190,7 @@ static void s_ecs_finalize_get_credentials_query(struct aws_credentials_provider
         .token_name = "Token",
         .account_id_name = "AccountId",
         .expiration_name = "Expiration",
-        .token_required = true,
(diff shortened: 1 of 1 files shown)


TingDaoK:
merged with https://github.com/awslabs/aws-c-auth/pull/292

PR #302, opened 2026-08-12
CI/CD improvements

azkrishpy:
*Issue #, if available:*

N/A

*Description of changes:*

Adds CI/CD workflows mirroring awslabs/aws-c-common#1250:
- `check-abi.yml` — ABI compliance check on PRs
- `release.yml` — manual (`workflow_dispatch`) release with auto version bump + tag
- `block-needs-review.yml` — gate merges labeled `needs-review`
- `VERSION` file seeded from latest release tag
- CMakeLists.txt: derive target `VERSION` from `VERSION` file via `aws_get_version()`


.github/workflows/release.yml
TingDaoK, line 1: somewhere, we need to document this. 
Will it be a concern if multiple developer try to kick off the action? 
can we do something to prevent the action to run when there is another one in progress? 
the abi check will take time, and it needs to build the package, right?
What if the build fails or some random failure goes on (eg: pull deps fails)?  
How will the dev find it out?

azkrishpy: > somewhere, we need to document this.

There is a guidance doc available for cutting releases. Will update it with some last minute stuff but the core idea already of auto releases already exists.

> Will it be a concern if multiple developer try to kick off the action?

Worst case scenario is two people try to run release at the same time, one person commits to main first and the other person's commit fails since we don't force push the branch. They would have to restart release.

ref: https://github.com/awslabs/aws-crt-builder/blob/0cb166b3cb3f3cee1b61e5139919e7d3a2adca51/.github/actions/auto-release/scripts/cut-release.sh#L94

> the abi check will take time, and it needs to build the package, right?
What if the build fails or some random failure goes on (eg: pull deps fails)?
How will the dev find it out?

Release job provides logs: https://github.com/awslabs/aws-c-common/actions/runs/31543641609
the abi-check action uses the builder package to build, so if the CI passes, we should not really have any issues building for the ABI check. (https://github.com/awslabs/aws-crt-builder/blob/0cb166b3cb3f3cee1b61e5139919e7d3a2adca51/.github/actions/check-abi/scripts/build.sh#L91).
If the ABI check fails, the release fails and the dev would have to look through the logs to investigate.

CMakeLists.txt
```diff
@@ -54,6 +54,7 @@ aws_check_headers(${PROJECT_NAME} ${AWS_AUTH_ROOT_HEADERS})
 aws_add_sanitizers(${PROJECT_NAME})
 
 # We are not ABI stable yet
+# TODO: switch to VERSION derived from the VERSION file (via aws_get_version) once SOVERSION is set explicitly, to avoid changing the SONAME.
 set_target_properties(${PROJECT_NAME} PROPERTIES VERSION 1.0.0)
```
sfod, line 58: Why not to implement the same logic as in [aws-c-common](https://github.com/awslabs/aws-c-common/pull/1250/changes#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR21)?

azkrishpy: Ideally I was hoping to do all of it after the first release. Common was already defining the SOVERSION as 1 and I kept that unchanged while changing version alone to ensure it was doing it correctly. Many of the others don't and seems like then Cmake would assume SOVERSION using the build version. Will update that as part of the last set of PRs so that we might break people only once.

sfod (approved):

VERSION
```diff
@@ -0,0 +1 @@
+0.10.3
```
sfod: Why is this one patch behind the current version?

azkrishpy: that is weird, will fix that

sfod (changes requested):

VERSION
sfod: From what I see the same happened in all repos.

sfod (approved):

.github/workflows/release.yml
TingDaoK: https://github.com/awslabs/aws-c-s3/blob/main/.github/workflows/ci.yml#L8-L11

we can add something like this to prevent concurrency for this job, just in case.

TingDaoK (approved):

